PR #2, opened 2016-04-13
RFC1: The Mediachain Datastore

vyzo:
The branch adds rfc/mediachain-rfc-1.md which contains the Mediachain Datastore specification in markdown.

Changes since last review:
DataModel:
- Cleaned up the signature and key structures to be consistent and carry references
- Added an EntityLinkCell to the entity chains so that we can have linkage between entities and support deduplication.

Structure:
- Added Overview, Example, and References sections

### DIFF
--- a/rfc/mediachain-rfc-1.md
+++ b/rfc/mediachain-rfc-1.md
@@ -82,7 +82,7 @@ We can make things more concrete with the following schema:
 Entity = {
  "type" : "entity"
  "name" : <String>
- [key : <Reference>]
+ ["key" : <Reference>]
  "signatures" : <Signatures>
  <Key> : <Value> ... ; entity metadata
  }
@@ -142,13 +142,16 @@ Nil = { "type" : "nil" }
 The entity and artefact chains differ on the type of relationships
 they store. They both support metadata update cells, which simply
 provide new metadata to the base object as arbitray key-value pairs.
+
 The entity chain also allows for linking two entities, for example
 having a pseudonymous artist publicly revealing a true name that
 points to another entity in the system.
+
 The artefact chain further stores links to entities that
 represent creation and potential Intellectual Property rights,
 references in the Internet media-space, and derivative relations
 to other artefacts.
+
 Each chain cell carries also carries one or more signatures by
 signatory entities in the system.
 
--- a/rfc/mediachain-rfc-1.md
+++ b/rfc/mediachain-rfc-1.md
@@ -58,7 +58,7 @@ date fields.
 Artefacts can also have associated their data stored in IPFS, so that media
 can be directly accessed from references to their Canonicals.
 If this is the case, then the artefact object will contain a link to
-the IPFS datablob in its `type` field
+the IPFS datablob in its `data` field
 
 Finally, both types carry a set of cryptographic signatures that assert their
 validity. The signatures come from _signatories_ with known keys in the
@@ -109,6 +109,12 @@ through linked lists. Thus for each entity and artefact in the
 system, we also have an associated chain which initially points to the
 canonical of a global `Nil` object:
 ```
+Chain = <ChainCell> | <Nil>
+ChainCell = {
+ "type" : ...
+ "chain" : <Reference>
+ <Key> : <Value> ...
+ }
 Nil = { "type" : "nil" }
 ```
 
@@ -134,8 +140,8 @@ EntityChainCell = <EntityUpdateCell> | <Nil>
 
 EntityUpdateCell = {
  "type" : "entityUpdate"
+ "chain" : <Reference>
  "entity" : <Reference>
- "entityChain" : <Reference>
  "signatures" : <Signatures>
   <Key> : <Value> ... ; metadata updates
  }
@@ -150,16 +156,16 @@ ArtefactChainCell =
  
 ArtefactUpdateCell = {
  "type" : "artefactUpdate"
+ "chain" : <Reference>
  "artefact" : <Reference>
- "artefactChain" : <Reference>
  "signatures" : <Signatures>
  <Key> : <Value> ... ; metadata updates
  }
 
 ArtefactCreationCell = {
  "type" : "artefactCreatedBy"
+ "chain" : <Reference>
  "artefact" : <Reference>
- "artefactChain" : <Reference>
  "entity" : <Reference>
  "signatures" : <Signatures>
  <Key> : <Value> ... ; creation metadata
@@ -167,17 +173,17 @@ ArtefactCreationCell = {
 
 ArtefactDerivationCell = {
  "type" : "artefactDerivedBy"
+ "chain" : <Reference>
  "artefact" : <Reference>
  "artefactOrigin" : <Reference>
- "artefactChain" : <Reference>
  "signatures" : <Signatures>
  <Key> : <Value> ... ; creation metadata
  }
 
 ArtefactOwnershipCell = {
  "type" : "artefactRightsOwnedBy"
+ "chain" : <Reference>
  "artefact" : <Reference>
- "artefactChain" : <Reference>
  "entity" : <Reference>
  "signatures" : <Signatures>
  <Key> : <Value> ... ; IP ownership metadata
@@ -185,8 +191,8 @@ ArtefactOwnershipCell = {
 
 ArtefactReferenceCell = {
  "type" : "artefactReferencedBy"
+ "chain" : <Reference>
  "artefact" : <Reference>
- "artefactChain" : <Reference>
  "url" : <URL>
  <Key> : <Value> ... ; reference metadata
 ```
--- a/rfc/mediachain-rfc-1.md
+++ b/rfc/mediachain-rfc-1.md
@@ -101,14 +101,11 @@ Reference = {
  "@link" = <Canonical>
  }
 
-Signatures = {
- "_1" : {
-  "entity" : <Reference>
-  "signature" : <IPRSSignature>
- }
- "_2" : ...
- ...
- }
+Signatures = [
+  { "entity" : <Reference>
+    "signature" : <IPRSSignature>
+  } ...
+ ]
 
 IPRSSignature = {
  "value" : <Reference>
--- a/rfc/mediachain-rfc-1.md
+++ b/rfc/mediachain-rfc-1.md
@@ -24,7 +24,7 @@ Open Data Peer-to-Peer network.
 
 The Mediachain is stored in a distributed append-only datastore with
 content addressable access. This maps well to the IPFS[2] peer-to-peer
-network, which names and accesses data using the SHA256 hash of
+network, which names and accesses data using the cryptographic hashes of
 the content. Objects stored in the network are immutable and persistent;
 their hashes become their _Canonical_ identifiers through which they can
 be referenced.
--- a/rfc/mediachain-rfc-1.md
+++ b/rfc/mediachain-rfc-1.md
@@ -11,13 +11,13 @@ Author: @vyzo
 ### Data Objects
 
 The Mediachain is stored in a distributed append-only datastore with
-content addressable access. This maps well to the IPFS peer-to-peer
+content addressable access. This maps well to the IPFS[2] peer-to-peer
 network, which names and accesses data using the SHA256 hash of
 the content. Objects stored in the network are immutable and persistent;
 their hashes become their _Canonical_ identifiers through which they can
 be referenced.
 
-Mediachain objects follow the IPLD[*] data model,
+Mediachain objects follow the IPLD[3] data model,
 allowing for storage of key-valued pairs where keys
 are strings and values are other key-valued pairs or strings.
 
@@ -90,8 +90,12 @@ Reference = {
  }
 
 Signatures = {
- "type" : "signatures"
- <SignatoryCanonical> : <IPRSSignature> ...
+ "_1" : {
+  "entity" : <Reference>
+  "signature" : <IPRSSignature>
+ }
+ "_2" : ...
+ ...
  }
 
 IPRSSignature = {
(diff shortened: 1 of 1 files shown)


rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+validity. The signatures come from _signatories_ with known keys in the
+system; any entity that contains a public key in its data can act
+as a signatory.
+
+We can make things more concrete with the following schema:
+```
+Entity = {
+ "type" : "entity"
+ "name" : <String>
+ ["key" : <Reference>]
```
bigs: the semantics of `"key"` are quite difficult to parse. is this a link to a public key? if so, why just one?

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+ "signatures" : <Signatures>
+ <Key> : <Value> ... ; entity metadata
```
bigs, line 87: should metadata be held in a compound value? this helps us prevent problems with duplicate keys, i.e. `name` being a piece of the metadata for some future object.

```
"metadata": <CompoundValue>
```

rfc/mediachain-rfc-1.md
vyzo: it is intended to provide public keys -- and it can be a keychain, not just a single key.
It points to a Reference, which is an IPFS blob which stores it in the system.

Perhaps change it to keychain?

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+ }
+
+Artefact = {
+ "type" : "artefact"
+ "name" : <String>
+ ["created" : <Date>]
+ ["description" : <String>]
+ ["data" : <Reference>]
+ "signatures" : <Signatures>
+ <Key> : <Value> ... ; artefact metadata
```
bigs, line 97: same here

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+ }
+
+Reference = {
+ "@link" = <Canonical>
+ }
+
+Signatures = {
+ "_1" : {
```
bigs: if these are integer indexed, could be an array. if u want to give them names, perhaps of the entity reference or something, i could see using a map

rfc/mediachain-rfc-1.md
vyzo, line 97: I have modified the wording and the field name to be keychain, so that it is explicit that we can have multiple set of keys in a standard blob format instead of doing our own storage.

rfc/mediachain-rfc-1.md
vyzo: We can't use arrays in IPLD, so I model it as a tuple with arbitrary length.

rfc/mediachain-rfc-1.md
vyzo, line 87: we have to, otherwise we can't do IPLD linkage.
Really, this is just the IPLD format with our schema and types on top.

rfc/mediachain-rfc-1.md
vyzo: changed to keychain, see note below.

rfc/mediachain-rfc-1.md
vyzo, line 97: also, here the <Key> datatype is defined above as a <String>.
The intent is to express that the metadata set is open, you can have multiple addtional field with some userland schema.

rfc/mediachain-rfc-1.md
parkan: What makes you say that we can't use arrays in IPLD?

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+
+EntityUpdateCell = {
+ "type" : "entityUpdate"
+ "chain" : <Reference>
+ "entity" : <Reference>
+ "signatures" : <Signatures>
+  <Key> : <Value> ... ; metadata updates
+ }
+
+EntityLinkCell = {
```
bigs, line 176: i think this is the most straightforward system, but it means you can't have links created in "transactions"--it will always require a round trip.

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+
+Note that the types of cells that can appear in either type of
+chain it is not limited to the types of relationships described
+above. These are merely the minimum schema to support the
+functionality of the Mediachain.
+
+Keeping this in mind, we can model the chains with the following
+schema:
+```
+EntityChainCell =
```
bigs, line 163: don't we need `EntityCreationCell`?

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+system from a small-scale proof of concept to a fully distributed,
+Open Data Peer-to-Peer network.
+
+## The Mediachain Data Model
+
+### Data Objects
+
+The Mediachain is stored in a distributed append-only datastore with
+content addressable access. This maps well to the IPFS[2] peer-to-peer
+network, which names and accesses data using the SHA256 hash of
```
parkan: I would maybe be less specific about the type of hash because the addresses are designed to be self-describing and support a variety of hash algos: https://github.com/ipfs/go-ipfs/tree/master/Godeps/_workspace/src/github.com/jbenet/go-multihash

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+ArtefactDerivationCell = {
+ "type" : "artefactDerivedBy"
+ "chain" : <Reference>
+ "artefact" : <Reference>
+ "artefactOrigin" : <Reference>
+ "signatures" : <Signatures>
+ <Key> : <Value> ... ; creation metadata
+ }
+
+ArtefactOwnershipCell = {
```
bigs, line 220: imo ownership is just a piece of metadata and should be handled in `ArtefactUpdateCell`

rfc/mediachain-rfc-1.md
vyzo: Perhaps my misunderstanding?
I was under the impression that IPLD is structure plus strings, no mention of arrays in the spec.
I am changing the Signatures type to be an array.

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+ArtefactCreationCell = {
+ "type" : "artefactCreatedBy"
+ "chain" : <Reference>
+ "artefact" : <Reference>
+ "entity" : <Reference>
+ "signatures" : <Signatures>
+ <Key> : <Value> ... ; creation metadata
+ }
+
+ArtefactDerivationCell = {
```
bigs, line 211: maybe this can lay dormant for now. i'm not sure we want to model derivative works yet

bigs, line 211: and if we do, i imagine it can be another piece of metadata that is actually a `Reference` :smile:

rfc/mediachain-rfc-1.md
parkan, line 163: I am guessing the first `EntityUpdateCell` is de facto `EntityCreationCell`?

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+ArtefactOwnershipCell = {
+ "type" : "artefactRightsOwnedBy"
+ "chain" : <Reference>
+ "artefact" : <Reference>
+ "entity" : <Reference>
+ "signatures" : <Signatures>
+ <Key> : <Value> ... ; IP ownership metadata
+ }
+
+ArtefactReferenceCell = {
```
bigs, line 229: same

rfc/mediachain-rfc-1.md
vyzo, line 176: Why not?
You can always add additional links in the rest of the metadata, any key can point to a Reference.

rfc/mediachain-rfc-1.md
parkan, line 163: Ah but there's an `ArtefactCreationCell`, this should be consistent, in either direction -- no preference

vyzo, line 163: no, Entity creations are tracked in the journal; they are root pointers representing canonical entry points in the Mediachain.

rfc/mediachain-rfc-1.md
vyzo: ok, changing to abstract "cryptographic hash"

rfc/mediachain-rfc-1.md
parkan, line 211: I read this as derivation of this set of statements from a raw piece of metadata, is that not correct?

Separately, we should model derivative works/"sameness" as per https://github.com/mediachain/L-SPACE/issues/28 but that's a separate/hard discussion

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+In order to make the mediachain indexable and discoverable we need to
+maintain a _Journal_ of updates to the data store. The journal is
+very similar to a blockchain in that it acts as a public ledger of all
+Mediachain data transactions in the system. By replaying the journal
+and fetching the data from IPFS, any node in the Internet can bootstrap
+an index that allows it to read and query the Mediachain.
+
+### Appending Data in the Mediachain
+
+The datastore initially is almost empty, containing only the `Nil` object
```
bigs, line 257: just thinking, one thing that is very cool about the `Nil` object is it only has one possible construction and will always hash to the same thing, meaning that regardless of whoever creates it, `Nil` is absolute!

rfc/mediachain-rfc-1.md
vyzo, line 220: I think it is an important enough item to warrant its own specified Cell type.

parkan, line 220: agreed, we may need some structure to make the licenses machine readable but that's outside the scope of this document and should be defined within an UpdateCell

rfc/mediachain-rfc-1.md
parkan, line 229: =bigs

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,432 @@
...
+JournalEntry =
+ <CanonicalEntry>
+ <ChainEntry>
+
+CanonicalEntry = {
+ "type" = "insert"
+ "ref"  = <Reference> ; canonical reference for an entity or artefact
+ }
+ 
+ChainEntry = {
```
parkan, line 281: maybe put a vector clock in here?

rfc/mediachain-rfc-1.md
vyzo, line 211: We can revisit specific cell once we have implemented enough of a spec to know how to tackle this.
We can always attach more specified fields based on how we crack this issue, it could contain a fingerprint from a suitable fingerprinting procedure depending on the media type.

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,435 @@
...
+The datastore initially is almost empty, containing only the `Nil` object
+as the bottom of all chains 
+For each entity and artefact added to the store, a corresponding entry
+is added to the journal, connecting a canonical reference with its chain
+pointing to the `Nil` object.
+Similarly, for every cell added to a chain, a corresponding entry is
+added to the journal, updating the head of the chain for a canonical
+reference.
+
+Thus the journal contents can be described as a sequence of entries
```
bigs, line 266: i'm not entirely sure why the journal exists... it seems to replicate all of the information stored in the chain

rfc/mediachain-rfc-1.md
vyzo, line 257: yup. The first thing will put on IPFS will be Nil and then we'll have its hash for ever :)

rfc/mediachain-rfc-1.md
vyzo, line 281: I thought about timestamping it, but I am not sure what sort of clock we should be keeping.
Let's revisit this field later when we have a better idea of how we'll do timestamps?

vyzo, line 281: Actually, on better thought I'll add timestamps -- I'll just wont specified the format for it.
When we have a better idea how the ledger protocol will work we can speicify it.

rfc/mediachain-rfc-1.md
parkan, line 163: what's the rationale for different behavior for Entity vs Artefact?

rfc/mediachain-rfc-1.md
bigs: https://github.com/jbenet/go-multihash :stuck_out_tongue:

rfc/mediachain-rfc-1.md
vyzo, line 281: ok, added a timestamp field for the journal entries.

rfc/mediachain-rfc-1.md
```diff
@@ -0,0 +1,432 @@
...
+pointer can act as a signatory.
+
+We can make things more concrete with the following schema:
+```
+Entity = {
+ "type" : "entity"
+ "name" : <String>
+ ["keychain" : <Reference>]
+ "signatures" : <Signatures>
+ <Key> : <Value> ... ; entity metadata
```
bigs, line 87: imo entitymetadata should still be a sub object in CBOR/JSON parlence...

```
"medatadata": {
  <Key> : <Value>
  ...
}
```

this is unambiguous and scalable

rfc/mediachain-rfc-1.md
bigs, line 176: the entities you're linking must already publish. you can't insert-and-link in a single step, which could be nice. not required, tho! seems reasonable the way it is

vyzo:
Ok so per our discussion at #tech I am merging into master.
It stays DRAFT until we implement and then we revisit based on our experience to push it to STANDARDS TRACK, before it becomes FINAL.
Gotta love ietf terminology.

PR #3, opened 2016-04-13
RFC1: Add the Journal for the example Mediachain

vyzo:
Adds the Journal entries for the example Mediachain. 
It also adds wording to make it clear that the journal itself can be implemented as a chain, but this decision is out of scope from RFC1 spec.


bigs:
LGTM :+1:

Issue #9, opened 2016-04-20
Update model


parkan:
This is an umbrella issue for discussing the update model.

In the L-SPACE design, each ImageBlob contains a complete set of translated statements from a piece of raw metadata. To obtain a "reconciled" view, we (theoretically) walk the chain and compare and combine the values. This way, two parties can make different (conflicting, i.e. I say this painting is from 1989, you say it's from 1990) statements frictionlessly at write time, then at read time we can decide which one wins through either a reputation system or an "arbitration" cell (I cede that I was wrong and you were right, or a respected art historian comes in and says you were wrong and I was right). This maps very nicely to a selective trust world, and captures the idea that contributing institutions' data is _never overwritten_, which is something we've promised to GLAMs.

In the RFC, the corresponding unit is the `ArtefactCreationCell`/`ArtefactUpdateCell`. The naming seems to imply (though the text doesn't say one way or another) that successive cells override the previous value. This makes the output easier to reason about (in a "one true view" world), but does not capture the properties of isolation and arbitrary read-time reconciliation that we want.

We should make an explicit decision about this behavior, and sketch out some approximate reconciliation rules

parkan:
assigned @vyzo here to make him aware of the discussion but really this is for everyone @bigs @yusefnapora @denisnazarov

vyzo:
Well, at the base level the interpretation and merging of the metadata is totally client-side, so there is no problem with having a merge by reputation. You are right however, the wording seems to be imply an override relation. We can add a paragraph that makes it explicit that the merge is open to the reader for interpretation.

A resolution by reputation is a good idea, but it still happens at the client-side.
I don't think we need a separate cell type for overriding metadata values other than an UpdateCell, and there is also nothing that forbids us from having multiple creation cells either!

So at a baseline, we can update to make it explicit that the metadata merge operation happens client-side (by readers of the mediachain, and this also includes services) and that multiple creation cells can appear in the chain.

vyzo:
While we are at it, maybe add an ArtefactLinkCell similar to the EntityLinkCell which can establish relationships other than derivation? Also, with such a cell it might make sense to absorb the derivation cell in the generic link cell.

Let's discuss.

vyzo:
I don't think we need to say anything about multiple creation cells, the algebraic datatype for ArtefactChainCell allows it.

parkan:
Agreed w/keeping things maximally generic -- we should have a separate discussion about broader datatypes

[...]

PR #13, opened 2016-04-21
rfc1: Add a paragraph making it clear that metadata merging is open

vyzo:
Adds a clarification paragraph for Issue #9


parkan:
Great!

PR #21, opened 2016-04-26
Journal Transactor: Copycat State Machine Implementation

vyzo:
#22

First pass on getting a state machine put together; compiles for now, and can muster some review.

Will need to add serialization support and tests.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
@@ -70,6 +70,12 @@ object Copycat {
     }
     
     private def updateChain(cmd: JournalUpdate): Xor[JournalError, ChainEntry] = {
+      def commit(ref: Reference, newchain: Reference, oldchain: Option[Reference]) = {
+        val entry = ChainEntry(nextSeqno(), ref, newchain, oldchain)
+        publishCommit(entry)
+        Xor.right(entry)
+      }
+      
       val ref = cmd.ref
       val cell = cmd.cell
       index.get(ref) match {
@@ -81,10 +87,7 @@ object Copycat {
                 val newcell = EntityChainCell(ref, chain, meta)
                 val newchain = datastore.put(newcell)
                 index.put(ref, EntityReference(Some(newchain)))
-                
-                val entry = ChainEntry(nextSeqno(), ref, newchain, chain)
-                publishCommit(entry)
-                Xor.right(entry)
+                commit(ref, newchain, chain)
               } else commitError("invalid chain cell")
             }
             case (ArtefactReference(chain), ArtefactChainCell(artefact, xchain, meta)) => {
@@ -92,10 +95,7 @@ object Copycat {
                 val newcell = ArtefactChainCell(ref, chain, meta)
                 val newchain = datastore.put(newcell)
                 index.put(ref, ArtefactReference(Some(newchain)))
-                
-                val entry = ChainEntry(nextSeqno(), ref, newchain, chain)
-                publishCommit(entry)
-                Xor.right(entry)
+                commit(ref, newchain, chain)
               } else commitError("invalid chain cell")
             }
             case _ => commitError("invalid chain")
@@ -103,7 +103,7 @@ object Copycat {
         }
       }
     }
-
+    
     def lookup(commit: Commit[JournalLookup]): Option[Reference] = {
       try {
         index.get(commit.operation.ref) match {
--- a/transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
@@ -75,39 +75,30 @@ object Copycat {
       index.get(ref) match {
         case None => commitError("invalid reference")
         case Some(cref) => {
-          cref match {
-            case EntityReference(chain) => {
-              cell match {
-                case EntityChainCell(entity, xchain, meta) => {
-                  if (checkUpdate(ref, chain, entity, xchain)) {
-                    val newcell = EntityChainCell(ref, chain, meta)
-                    val newchain = datastore.put(newcell)
-                    index.put(ref, EntityReference(Some(newchain)))
-                    
-                    val entry = ChainEntry(nextSeqno(), ref, newchain, chain)
-                    publishCommit(entry)
-                    Xor.right(entry)
-                  } else commitError("invalid chain cell")
-                }
-                case _ => commitError("invalid chain")
-              }
+          (cref, cell) match {
+            case (EntityReference(chain), EntityChainCell(entity, xchain, meta)) => {
+              if (checkUpdate(ref, chain, entity, xchain)) {
+                val newcell = EntityChainCell(ref, chain, meta)
+                val newchain = datastore.put(newcell)
+                index.put(ref, EntityReference(Some(newchain)))
+                
+                val entry = ChainEntry(nextSeqno(), ref, newchain, chain)
+                publishCommit(entry)
+                Xor.right(entry)
+              } else commitError("invalid chain cell")
             }
-            case ArtefactReference(chain) => {
-              cell match {
-                case ArtefactChainCell(artefact, xchain, meta) => {
-                  if (checkUpdate(ref, chain, artefact, xchain)) {
-                    val newcell = ArtefactChainCell(ref, chain, meta)
-                    val newchain = datastore.put(newcell)
-                    index.put(ref, ArtefactReference(Some(newchain)))
-                    
-                    val entry = ChainEntry(nextSeqno(), ref, newchain, chain)
-                    publishCommit(entry)
-                    Xor.right(entry)
-                  } else commitError("invalid chain cell")
-                }
-                case _ => commitError("invalid chain")
-              }
+            case (ArtefactReference(chain), ArtefactChainCell(artefact, xchain, meta)) => {
+              if (checkUpdate(ref, chain, artefact, xchain)) {
+                val newcell = ArtefactChainCell(ref, chain, meta)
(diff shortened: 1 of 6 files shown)


transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+    cell: ChainCell
+  ) extends Command[Xor[JournalError, ChainEntry]]
+  
+  case class JournalLookup(
+    ref: Reference
+  ) extends Query[Option[Reference]]
+  
+  case class JournalCommitEvent(entry: JournalEntry)
+
+  class JournalStateMachine(
```
bigs: this probably warrants its own source file as it's got a bit of logic to it

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+    val datastore: Datastore
+  ) extends StateMachine with Snapshottable with SessionListener {
+    sealed abstract class CanonicalReference {
```
bigs: this can be a `sealed trait`. i think u only need to use `sealed abstract class` when u need to accept type params

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+    }
+    case class EntityReference(chain: Option[Reference]) extends CanonicalReference
+    case class ArtefactReference(chain: Option[Reference]) extends CanonicalReference
+    
+    private var seqno: BigInt = 0
+    private var index: MMap[Reference, CanonicalReference] = new MHashMap    
+    private var clients: MSet[ServerSession] = new MHashSet // this wanted to be called sessions
+
+    
+    private def CommitError(what: String) = Xor.left(JournalCommitError(what))
```
bigs: `commitError`

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+    val datastore: Datastore
+  ) extends StateMachine with Snapshottable with SessionListener {
+    sealed abstract class CanonicalReference {
+      def chain: Option[Reference]
+    }
+    case class EntityReference(chain: Option[Reference]) extends CanonicalReference
+    case class ArtefactReference(chain: Option[Reference]) extends CanonicalReference
+    
+    private var seqno: BigInt = 0
+    private var index: MMap[Reference, CanonicalReference] = new MHashMap
```
bigs: this & clients can be `val`s because they hold mutable objects. ref to root object shouldn't change

bigs: `val` = single assignment

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+        insertRecord(commit.operation)
+      } finally {
+        commit.release()
+      }
+    }
+    
+    private def insertRecord(cmd: JournalInsert): Xor[JournalError, CanonicalEntry] = {
+      val rec = cmd.record
+      val ref = datastore.put(rec)
+      index.get(ref) match {
```
parkan: I would probably write this as

```
val existing = index.get(ref)
(existing, rec) match {
  case (Some(_), _) => CommitError...
  case (None, Entity(_)) => index += EntityReference(None)
  case (None, Artefact(_)) => ...
}
```

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+        updateChain(commit.operation)
+      } finally {
+        commit.release()
+      }
+    }
+    
+    private def updateChain(cmd: JournalUpdate): Xor[JournalError, ChainEntry] = {
+      val ref = cmd.ref
+      val cell = cmd.cell
+      index.get(ref) match {
```
parkan: Likewise, I would try to write this in terms of tuples instead of nesting, since the branches are quite unbalanced.

You can also embed your `if` into the case statement as a guard, see https://www.safaribooksonline.com/library/view/scala-cookbook/9781449340292/ch03s14.html

Scala is all about pattern matching!

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+                    
+                    val entry = ChainEntry(nextSeqno(), ref, newchain, chain)
+                    publishCommit(entry)
+                    Xor.right(entry)
+                  } else CommitError("invalid chain cell")
+                }
+                case _ => CommitError("invalid chain")
+              }
+            }
+            case ArtefactReference(chain) => {
```
parkan: Here

```
case (ArtefactReference(chain), ArtefactChainCell(...)) => ...
case (ArtefactReference(_), _) => CommitError("invalid chain")
```

would immediately tell us why the error occurred (mismatched reference and cell)

transactor/src/main/scala/io/mediachain/transactor/Types.scala
```diff
@@ -69,5 +69,14 @@ object Types {
...
-  abstract class JournalError
+  sealed abstract class JournalError
+  
+  case class JournalCommitError(what: String) extends JournalError {
+    override def toString = "Journal Commit Error: " + what
+  }
+  
+  // Datastore interface
+  trait Datastore {
+    def put(rec: Record): Reference
```
bigs, line 105: we need more than `put`. i think we need at least 

``` scala
def head(): Reference
```

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
```diff
@@ -0,0 +1,186 @@
...
+    }
+    
+    
+    private def checkUpdate(ref: Reference, chain: Option[Reference],
+                            xref: Reference, xchain: Option[Reference]) = {
+      (xref == ref) && ((xchain == None) || (xchain == chain))
+    }
+        
+    // Snapshottable
+    override def install(reader: SnapshotReader) {
```
parkan: I think you don't need to `override` these because you're implementing an abstract interface, but not 100% how this works with java interop

parkan: On second thought, maybe leave the `override` in there so the compiler complains if the parent interface changes

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
parkan: both traits and classes can accept type params, though classes can accept constructor params and can be called from java

http://www.artima.com/pins1ed/traits.html#12.7

I would err on the side of traits most of the time, though

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
vyzo: these are the state machine messages, i think they belong in this module

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
vyzo: They need to be loadable by the Snapshot, hence vars.

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
vyzo: @bigs extracted the type and normalized a bit, there is a single index += (ref -> rec.reference) now.

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
vyzo: yeah, the override is useful for verifying that the signature matches for me.

vyzo:
Dropped the parens on CanonicalRecord.reference calls as there are no side-effects

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
vyzo: yeah, rewriting to take advantage of tuple patters, the code drifts less to the right this way.

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
vyzo: oh you mean having the state machine on its own file instead of Copycat?
Probably a good idea -- I'll refactor once I have some more code to put in the different files.

vyzo: refactored the module to StateMachine in the branch's tip.

transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
vyzo: clients can be val, I turned it into one in the tip.

vyzo:
So in the latest commit I added dummies for store and reference and did some simple testing in the sbt console.
You can try out the state machine with something like this:

``` scala
import io.mediachain.transactor._
import io.atomix.catalyst.transport.Address
val dummystore = new Dummies.DummyStore
val server = Copycat.Server.build(new Address("127.0.0.1:10000"), "/tmp/copycat/log0", dummystore)
server.bootstrap().join()
val client = Copycat.Client.build()
client.connect(new Address("127.0.0.1:10000")).join()
val res = client.submit(StateMachine.JournalInsert(Types.Entity(Map()))).join()
```

vyzo:
renamed the poorly named CanonicalReference to ChainReference to make it clearer what the intent of its hierarchy is.

PR #27, opened 2016-04-26
Bigs WIP

bigs:
Small changes here, @vyzo. Some broader comments:
- I'm not really sure what the `CanonicalReferences` are. Are these IDs?
- With some type parameterization, we can reduce the size of `updateChain` by half i.e. have `ChainCell[T]` and `CanonicalReference[T]` where `T` can be `Artefact` or `Entity` etc
- I don't really understand the purpose of `index`. We need commenting on these types. Canonical is an overloaded term in our world -- it refers to the external identifier end users can use to refer to a work. Beyond that name clash, I'm just not sure why this was the data structure chosen. Would love to see an impl of `head` as well.

Maybe we can work out a process where I send you PRs at the end of my day

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/Copycat.scala
@@ -31,18 +31,11 @@ object Copycat {
   class JournalStateMachine(
     val datastore: Datastore
   ) extends StateMachine with Snapshottable with SessionListener {
-    sealed abstract class CanonicalReference {
-      def chain: Option[Reference]
-    }
-    case class EntityReference(chain: Option[Reference]) extends CanonicalReference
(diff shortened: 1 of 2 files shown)


bigs:
@vyzo go ahead and review/merge

vyzo:
I added the CanonicalReference types as a mean for the state machine to distinguish between artefact and entity chains (so that we dont cons an EntityChainCell on an Artefact chain, etc).

The index is the mapping of references (artefact or entity) to chain heads, what the state machine effectively computes.
I can add some comment on it or give it a better name.

vyzo:
Up to you to delete the branch, dunno if you are still working on it!

vyzo:
I refactored the Chain construction code and commit publish, and now it is much denser - no type parameterization needed, rewriting for a generic type would be bigger than the use.

PR #31, opened 2016-04-27
JSON Serializers for transactor types

yusefnapora:
This adds a `JsonSerialization` object that contains converters to/from the transactor types and json4s `JValue`s.

It also creates copycat `TypeSerializer` classes for each serializable type, which are registered with the client and server's `Serializer`.  

At the moment the serialized json is rendered to a UTF-8 string, but we could use CBOR by pulling in the CBOR code from L-SPACE

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/JsonSerialization.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/JsonSerialization.scala
@@ -69,9 +69,7 @@ object JsonSerialization {
   } + referenceSerializer
 
 
-  // Below are conversion functions from transactor types to
-  // json4s `JObject`s.  We may want to write a better conversion
-  // interface/
+  // Conversion functions from transactor types to json4s `JObject`s.
 
   def toJObject(record: Record): JObject =
     Extraction.decompose(record).asInstanceOf[JObject]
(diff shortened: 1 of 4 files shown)


vyzo:
JSON might be even worse than pojo in serialization efficiency

transactor/src/main/scala/io/mediachain/transactor/JsonSerialization.scala
```diff
@@ -0,0 +1,64 @@
+package io.mediachain.transactor
+
+import io.mediachain.transactor.Types._
+import org.json4s._
+
+case class IPLDReference(link: String) extends Reference
```
vyzo: this should have an internal binary representation for storage efficiency reasons

yusefnapora: yeah, I think I'll remove this type from this branch entirely.  I was using it before you merged in the DummyReference, which will do for now.  We can more fully work out the IPFS references when we tackle the IPFS datastore

transactor/src/main/scala/io/mediachain/transactor/Types.scala
```diff
@@ -5,15 +5,15 @@ object Types {
   import org.json4s.JValue
 
   // Mediachain Datastore Records
-  sealed abstract class Record extends Serializable {
```
vyzo, line 8: don't remove this, we want to retain the ability to run with pojo

transactor/src/main/scala/io/mediachain/transactor/Types.scala
```diff
@@ -5,15 +5,15 @@ object Types {
...
+  sealed abstract class Record {
     def meta: Map[String, JValue]
   }
   
   // References to records in the underlying datastore
-  abstract class Reference extends Serializable
```
vyzo, line 13: same here

transactor/src/main/scala/io/mediachain/transactor/Types.scala
```diff
@@ -5,15 +5,15 @@ object Types {
...
+  abstract class Reference
 
   // Typed References for tracking chain heads in the StateMachine
-  sealed abstract class ChainReference extends Serializable {
```
vyzo, line 16: ditto

transactor/src/main/scala/io/mediachain/transactor/JsonSerialization.scala
```diff
@@ -0,0 +1,177 @@
...
+import org.json4s._
+import org.json4s.jackson.JsonMethods.{compact, render, parse}
+
+/**
+  * Represents a 'multihash' link to an IPFS resource
+  *
+  * @param multihash - the base58 encoded string representation of an IPFS
+  *             multihash
+  */
+case class IPFSReference(multihash: String) extends Reference
```
vyzo, line 16: as commented on the earlier commit -- this should have a binary internal representation, because Size Does Matter in this case.

parkan:
My assumption is that JValue are a reasonably efficient representation if we don't end up emitting actual JSON. JValue -> CBOR is something we'll need anyway for interacting with IPLD

transactor/src/main/scala/io/mediachain/transactor/JsonSerialization.scala
```diff
@@ -0,0 +1,177 @@
...
+  // Conversion functions from transactor types to json4s `JObject`s.
+
+  def toJObject(record: Record): JObject =
+    Extraction.decompose(record).asInstanceOf[JObject]
+
+  def toJObject(chainCell: ChainCell): JObject =
+    Extraction.decompose(chainCell).asInstanceOf[JObject]
+
+  def toJObject(ref: Reference): JObject =
+    Extraction.decompose(ref).asInstanceOf[JObject]
```
vyzo, line 94: you want a {"@link" = string-repr-of-reference} generated for this, how is this accomplished?

transactor/src/main/scala/io/mediachain/transactor/JsonSerialization.scala
```diff
@@ -0,0 +1,177 @@
...
+    override val typeHintFieldName: String = "type"
+  } + referenceSerializer
+
+
+  // Conversion functions from transactor types to json4s `JObject`s.
+
+  def toJObject(record: Record): JObject =
+    Extraction.decompose(record).asInstanceOf[JObject]
+
+  def toJObject(chainCell: ChainCell): JObject =
```
vyzo, line 90: here there is the intricacy of the chain pointer representation -- None must still generate an {"@iink"=...} pointer to the Nil object.

transactor/src/main/scala/io/mediachain/transactor/JsonSerialization.scala
parkan, line 16: agreed on this, we should have a multihash that's a bytestring

I would say let's alias `Multihash` as `String` for now and then swap the underlying later?

PR #32, opened 2016-04-27
Add simple serialization for datatypes

bigs:
@vyzo more work inc on actually serializing to bytes, but this is a start

### DIFF
--- a/transactor/build.sbt
+++ b/transactor/build.sbt
@@ -2,9 +2,10 @@ name := "transactor"
 scalaVersion := "2.11.7"
 
 libraryDependencies ++= Seq(
- "io.atomix.copycat" % "copycat-server" % "1.0.0-rc7",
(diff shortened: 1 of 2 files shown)


vyzo:
This approach is reasonable, and handles Artefacts and Entities fine.
I guess that for the chian cells we create subclasses that have the correct CBOR type set -- we can add the subclasses in Types, otherwise they will have to live at the interface barrier.

vyzo:
It is a little problematic with the Chain cells, as the Index constructs new instances directly to add the chain pointer on top of the metadata and the artefact/entity reference.
Perhaps we can have a constructor method in ChainCells to create the appropriate type from the subtypes.

vyzo:
I reverted the merge to let you work out the rest of the details, don't kill the branch!

PR #34, opened 2016-04-28
Test scaffolding and spec for JournalStateMachine index

vyzo:
Adds test scaffolding and a spec for the index invariants of the JournalStateMachine

### DIFF
--- a/transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
+++ b/transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
@@ -1,0 +1,44 @@
+package io.mediachain.transactor
+
+import io.atomix.copycat.client.CopycatClient
+import io.atomix.copycat.server.CopycatServer
+import io.atomix.catalyst.transport.Address
+
+case class DummyContext(
+  server: CopycatServer, 
+  client: CopycatClient, 
+  store: Types.Datastore,
+  logdir: String
+)
+
+object DummyContext {
+  def setup(srvaddr: String, logdir: String) = {
+    println("*** SETUP DUMMY COPYCAT CONTEXT")
+    setupLogdir(logdir)
+    val address = new Address(srvaddr)
+    val store = new Dummies.DummyStore
+    val server = Copycat.Server.build(address, logdir, store)
+    server.bootstrap().join()
+    val client = Copycat.Client.build()
+    client.connect(address).join()
+    DummyContext(server, client, store, logdir)
+  }
+  
+  def setupLogdir(logdir: String) {
+    import sys.process._
+    (s"rm -rf $logdir").!
+    (s"mkdir -p $logdir").!
+  }
+  
+  def cleanupLogdir(logdir: String) {
+    import sys.process._
+    (s"rm -rf $logdir").!
+  }
+  
+  def shutdown(context: DummyContext) {
+    println("*** SHUT DOWN DUMMY COPYCAT CONTEXT")
+    context.client.close().join()
+    context.server.shutdown().join()
+    cleanupLogdir(context.logdir)
+  }
+}
--- a/transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
+++ b/transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
@@ -12,9 +12,9 @@ case class DummyContext(
 )
 
 object DummyContext {
-  def setup(srvaddr: String, logdir: String) = {
+  def setup(srvaddr: String) = {
     println("*** SETUP DUMMY COPYCAT CONTEXT")
-    setupLogdir(logdir)
+    val logdir = setupLogdir()
     val address = new Address(srvaddr)
     val store = new Dummies.DummyStore
     val server = Copycat.Server.build(address, logdir, store)
@@ -24,10 +24,11 @@ object DummyContext {
     DummyContext(server, client, store, logdir)
   }
   
-  def setupLogdir(logdir: String) {
+  def setupLogdir() = {
     import sys.process._
-    (s"rm -rf $logdir").!
+    val logdir = "mktemp -d".!!
     (s"mkdir -p $logdir").!
+    logdir
   }
   
   def cleanupLogdir(logdir: String) {
--- a/transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
+++ b/transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
@@ -1,0 +1,117 @@
+package io.mediachain.transactor
+
+import org.specs2.specification.{AfterAll, BeforeAll}
+
+import java.util.function.Consumer
+import java.util.concurrent.{BlockingQueue, LinkedBlockingQueue, TimeUnit}
+
+import io.atomix.copycat.client.CopycatClient
+import io.atomix.catalyst.transport.Address
+
+import Types._
+import StateMachine._
+
+object JournalCommitSpec extends io.mediachain.BaseSpec
+  with BeforeAll
+  with AfterAll
+{
+
+  def is =
+    sequential ^
+  s2"""
+  JournalStateMachine publishes commit events:
+   - when inserting an entity $insertEntity
+   - when extending the entity chain $extendEntityChain
+   - when extending the entity chain further $extendEntityChainFurther
+   - when inserting an artefact $insertArtefact
+   - when extending the artefact chain $extendArtefactChain
+  """
+
+  def beforeAll() {
+    JournalCommitSpecContext.setup()
+  }
+  
+  def afterAll() {
+    JournalCommitSpecContext.shutdown()
+  }
+  
+  def insertEntity = {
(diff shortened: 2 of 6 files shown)


transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+class JournalCommitSpecContext(val dummy: DummyContext, 
+                               val qclient: CopycatClient, 
+                               val queue: BlockingQueue[JournalEntry]) {
+  var ref: Reference = null
+}
+
+object JournalCommitSpecContext {
+  var instance: JournalCommitSpecContext = null
+  def setup() {
+    val dummy = DummyContext.setup("127.0.0.1:10001", "/tmp/transactor-test/commit")
```
bigs: imo the directory should be created via `mktemp -d`

vyzo: good point, will fix.

transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
```diff
@@ -0,0 +1,44 @@
...
+
+case class DummyContext(
+  server: CopycatServer, 
+  client: CopycatClient, 
+  store: Types.Datastore,
+  logdir: String
+)
+
+object DummyContext {
+  def setup(srvaddr: String, logdir: String) = {
```
bigs: logdir should probably be generated in this function as mentioned earlier

vyzo: right, will disappear as argument if it is created with mktemp.

transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
```diff
@@ -0,0 +1,44 @@
...
+    println("*** SETUP DUMMY COPYCAT CONTEXT")
+    setupLogdir(logdir)
+    val address = new Address(srvaddr)
+    val store = new Dummies.DummyStore
+    val server = Copycat.Server.build(address, logdir, store)
+    server.bootstrap().join()
```
bigs, line 21: does `join` accept a timeout? if so, consider it so tests will eventually fail if they can't establish a context

vyzo, line 21: dunno, will check.

transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+  }
+  
+  def afterAll() {
+    JournalCommitSpecContext.shutdown()
+  }
+  
+  def insertEntity = {
+    val context = JournalCommitSpecContext.context
+    val res = context.dummy.client.submit(JournalInsert(Entity(Map()))).join()
+    res.foreach((entry: CanonicalEntry) => {context.ref = entry.ref})
```
bigs, line 41: some comments as to why u need to cache this would be useful

vyzo, line 41: you mean the reference or the context?

transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+  
+  def extendEntityChainFurther = {
+    val context = JournalCommitSpecContext.context
+    val ref = context.ref
+    val res = context.dummy.client.submit(JournalUpdate(ref, EntityChainCell(ref, None, Map()))).join()
+    val entry = context.queue.poll(1, TimeUnit.SECONDS)
+    res must beRightXor { (xentry: ChainEntry) =>
+      entry must_== xentry
+    }
+  }
```
bigs, line 66: might be nice to have a test verifying chain ordering

transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+
+class JournalCommitSpecContext(val dummy: DummyContext, 
+                               val qclient: CopycatClient, 
+                               val queue: BlockingQueue[JournalEntry]) {
+  var ref: Reference = null
+}
+
+object JournalCommitSpecContext {
+  var instance: JournalCommitSpecContext = null
+  def setup() {
```
bigs:
``` scala
def setup(): Unit = {
```

transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
vyzo, line 21: no timeout argument in join, we have to bite it -- but I think it fails eventually anyway by contract.

bigs:
few minor suggestions, otherwise LGTM :+1:

transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
vyzo, line 66: chain ordering is verified in the JournalIndex test -- here I am verifying that the journal commit events are received by a second client.

transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
vyzo: ok, will fix

transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
parkan, line 41: I'm not entirely following what's happening here

vyzo, line 41: basically I am threading context between the various sequential test calls, using the context object to store references returned by the state machine.
The cached references are used in subsequent calls to check the integrity of the state machine's response.

parkan, line 41: Ok, I see `res` is an `Xor` and you're setting JournalCommitSpecContext's `ref` to that result and then adding to it in `extendEntityChain` etc. Can we express this w/o side effects, in more of a unit style? I think it's preferable to have each test set up its fixtures and test an action in isolation rather than statefully chaining them in this way. We can also separately have an integration test that runs all of the actions and checks that the final state is sane.

parkan, line 41: Ok, after further thought, I think the reason this whole approach feels weird to me is that it's neither unit nor integration, but kind of inbetween. I think we can express the things you want here in a better way; I would:
- set up the context (w/o the extra initializer as per #37)
- in BeforeAll, perform all your updates, receiving the updates in a queue (queue and DummyStore are your state accumulators, don't need any vars)
- in one test, drain the queue and make sure it contains all the objects you expect (you can use a matcher like `Seq(1, 2, 3, 4) must contain(allOf(2, 4))`)
- in another test, request the chain for the entity and make sure it has the right objects in the right order
- etc

I think this will make a fine integration test. Separately, we can write unit tests that test methods like `JournalStateMachine.update` in isolation, without involving clients

vyzo, line 41: I think this is like standing on your head to satisfy some abstract notion of test structure :)
It is much simpler to write in sequential style with context threading.

vyzo, line 41: Also, the tests will still need to share context in the suggested approach: we need to references from the canonical record inserts in order to check their chain heads!

vyzo, line 41: I think part of the confusion stems from the kind of tests we need for a stateful component like the StateMachine.
The tests are written as functional specification acceptance tests, and not unit tests in the pure sense. 
I don't think that pure unit test style is appropriate for this part of the system's test suite.

parkan, line 41: The idea is to generate all your state and then run specifications against it. If you set everything up beforehand, the tests can be pure. I'm surprised I'm the one advocating purity and immutability over here! 😛  In Scala, using `var` is typically considered an absolute last resort.

I agree that unit testing is not necessarily the right thing here (at least until we start swapping out copycat components) but I definitely find this state passing hard to reason about. You don't have to spend time rewriting this, but please try to write side effect-free tests in the future.

vyzo, line 41: But the specification is exactly the threading and sequence of commands!
This is not a Unit test, confining all access to a simple object -- this is a functional specification for a sequence of operations and the results define the interface of the state machine.

I am definitely for side-effect free tests where possible, but this is just not possible when testing the behavior of a statefull component across a sequence of operations.

PR #35, opened 2016-04-28
Add CBOR methods from L-SPACE

yusefnapora:
This just pulls in the `CborMethods` from the L-SPACE project, and the `SortedObjectSerializer` which we use to ensure consistent ordering of map keys.

You should be able to do this:

``` scala
import io.mediachain.util.cbor.CborMethods.{bytes, render}

val cborBytes = bytes(render(someJValue))
```

to get the binary CBOR representation of any JValue.

Unfortunately, since json (and by extension, `JValue`) has no binary value type, we can't use this to store e.g. the multihash for a `@link` field as a byte array.  In theory we could special case that in the serializer, but it would be a pretty ugly hack.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -10,7 +10,7 @@ object MediachainBuild extends Build {
     scalaVersion := "2.11.7",
     libraryDependencies ++= Seq(
       "org.typelevel" %% "cats" % "0.4.1",
-      "org.json4s" %% "json4s-jackson" % "3.2.11"
+      "org.json4s" %% "json4s-jackson" % "3.3.0"
     )
   )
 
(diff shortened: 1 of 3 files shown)


vyzo:
LGTM, although @bigs is probably the more interested client :)

yusefnapora:
BTW, I've been reading through the code for this project: https://github.com/c-rack/cbor-java

It's pretty awesome...  It automatically sorts map keys according to the ["canonical CBOR" spec](https://tools.ietf.org/html/rfc7049#section-3.9), will pack integer and floating point values into the smallest possible type (as required by the canonical spec), and is very well-tested.

It might be worth using if we think that having support for binary values is important.  It would also be much simpler to eventually be compatible with the [IPLD canonical format](https://github.com/ipfs/specs/tree/master/ipld#serialised-cbor-with-tags), which uses tags to indicate a `@link` field.  

Converting from `JValue` to the `DataItem` type in cbor-java seems straightforward, although the `DataItem` to `JValue` conversion would obviously only work for json types (no byte strings, etc).

parkan:
this is also abandoned, I'm assuming

PR #36, opened 2016-04-28
alternative CBOR encoding

yusefnapora:
This is an alternative to #35 that doesn't use the jackson CBOR module.  Instead it uses [cbor-java](https://github.com/c-rack/cbor-java) and provides a `jValueToCbor` method to convert from JValue.  

Doesn't handle the other direction (cbor to json), since that's a lossy conversion.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
@@ -14,9 +14,8 @@ object Dummies {
     override def hashCode = num
     override def toString = "dummy@" + num
 
-    val CBORType = "dummyReference"
-    override def toCbor =
-      toCMapWithDefaults(Map("@link" -> CLong(num)), Map())
+    val CBORType = "" // unused
+    override def toCbor = CMap.withStringKeys("@link" -> CString(this.toString))
   }
   
   class DummyStore extends Datastore {
--- a/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
@@ -4,6 +4,7 @@ import scala.collection.mutable.{Map => MMap, HashMap => MHashMap}
 
 object Dummies {
   import io.mediachain.transactor.Types._
+  import io.mediachain.util.cbor._
   
   class DummyReference(val num: Int) extends Reference {
     override def equals(that: Any) = {
@@ -12,6 +13,10 @@ object Dummies {
     }
     override def hashCode = num
     override def toString = "dummy@" + num
+
+    val CBORType = "dummyReference"
+    override def toCbor =
+      toCMapWithDefaults(Map("@link" -> CLong(num)), Map())
   }
   
   class DummyStore extends Datastore {
--- a/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
@@ -4,7 +4,7 @@ import scala.collection.mutable.{Map => MMap, HashMap => MHashMap}
 
 object Dummies {
   import io.mediachain.transactor.Types._
-  import io.mediachain.util.cbor._
+  import io.mediachain.util.cbor.CborAST._
   
   class DummyReference(val num: Int) extends Reference {
     override def equals(that: Any) = {
--- a/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
@@ -16,12 +16,12 @@ object Dummies {
   
   class DummyStore extends Datastore {
     var seqno = 0
-    val records: MMap[Reference, Record] = new MHashMap
+    val store: MMap[Reference, DataObject] = new MHashMap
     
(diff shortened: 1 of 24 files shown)


transactor/src/main/scala/io/mediachain/util/cbor/JValueConversions.scala
```diff
@@ -0,0 +1,64 @@
...
+object JValueConversions {
+  import scala.language.implicitConversions
+  import collection.JavaConverters._
+
+  private def firstItem(builder: CborBuilder): Cbor.DataItem =
+    builder.build().asScala.headOption.getOrElse {
+    throw new MappingException("Unable to convert JValue to CBOR")
+  }
+
+  implicit def jValueToCbor(jValue: JValue): Cbor.DataItem =
```
yusefnapora: hmm, this should probably not be implicit...

transactor/src/main/scala/io/mediachain/util/cbor/JValueConversions.scala
```diff
@@ -0,0 +1,62 @@
...
+      case JInt(num) => firstItem {
+        new CborBuilder().add(num.bigInteger)
+      }
+
+      case JLong(num) => firstItem {
+        new CborBuilder().add(num)
+      }
+
+      case JDecimal(num) => firstItem {
+        new CborBuilder().add(num.doubleValue())
```
parkan: extra parens for non-sideeffecting accessor

bigs:
oooooooooo i like this! do we trust the lib? if so, go ahead and merge

vyzo:
it has conflicts with the changes in types for the blockchain.

transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
```diff
@@ -12,6 +13,10 @@ object Dummies {
     }
     override def hashCode = num
     override def toString = "dummy@" + num
+
+    val CBORType = "dummyReference"
```
parkan: this is to be replaced with the real type when it's reserved or?

yusefnapora: I think we'll have to do some more work to get links encoded properly once the tag is reserved.  

[The spec](https://github.com/ipfs/specs/tree/master/ipld#serialised-cbor-with-tags) basically says if there are no properties attached to the `{"@link": "..."}` object apart from the link itself, then you just write out the link value wrapped in the tag.  

But if you do have properties, you write out an array whose first item is the tagged link and whose second item is a map with the remaining properties.

So the complexity sort of depends on whether we want to support properties in the `@link` object.  Either way, we probably don't want to actually write out the `type` field... But this seemed like a decent way to go for now.

vyzo: Our usage in RFC has the references with a naked "@link" without properties.

parkan: I think it's ok to only support naked for now, things still seem not 100% shaken out in IPLD land

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
```diff
@@ -12,6 +13,8 @@ import cats.data.Xor
 
 object StateMachine {
   import io.mediachain.transactor.Types._
+
+  val JournalBlockSize: Int = 4096 // blocksize for Journal Blockchain
```
parkan, line 17: shouldn't be capitalCCd since it's not a type

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
```diff
@@ -25,15 +28,23 @@ object StateMachine {
   case class JournalLookup(
     ref: Reference
   ) extends Query[Option[Reference]]
+
+  case class JournalCurrentBlock() extends Query[JournalBlock]
   
-  case class JournalCommitEvent(entry: JournalEntry) extends Serializable
+  sealed abstract class JournalEvent extends Serializable
```
bigs, line 34: does this want `CborSerializable` too?

yusefnapora, line 34: ah yeah, it probably should be.  I haven't dug into the StateMachine types yet

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
vyzo, line 17: Let this be here, as it is out of scope for this branch. 
I'll fix it in a separate branch as it has dependent changes in the tests too.

transactor/src/main/scala/io/mediachain/transactor/TypeSerialization.scala
```diff
@@ -0,0 +1,243 @@
...
+
+
+object TypeSerialization {
+
+  import cats.data.Xor
+  import io.mediachain.transactor.Types._
+  import io.mediachain.util.cbor.CborAST._
+  import io.mediachain.util.cbor.CborCodec
+
+  object CBORTypeNames {
```
bigs, line 16: ahhhhhhhhh this is a good spot for this. :+1:

yusefnapora:
closing to reopen against master

PR #37, opened 2016-04-28
Remove unneccessary setup method and var

parkan:
I'm seeing many examples of this pattern where we start with a null var and then call an initializer method. This is really not necessary unless we want very precise control over when the init is called.

Scala objects (singletons) are already lazy, and have inline constructors:

``` scala
scala> object SoLazy {
     | println("executing body!")
     | val foo = { println("executing val!"); 99 }
     | def bar = { println("executing def!"); 22 }
     | def fire = 101
     | }
defined object SoLazy

scala> SoLazy.fire
executing body!
executing val!
res2: Int = 101
```

I might be missing something about your intention, but from my perspective this achieves the same effect. 

(this is not a complete set of changes but I wanted to isolate them to be illustrative)

### DIFF
--- a/transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
+++ b/transactor/src/test/scala/io/mediachain/transactor/JournalCommitSpec.scala
@@ -28,7 +28,6 @@ object JournalCommitSpec extends io.mediachain.BaseSpec
   """
 
   def beforeAll() {
-    JournalCommitSpecContext.setup()
   }
   
   def afterAll() {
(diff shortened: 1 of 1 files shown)


vyzo:
The setup is expensive and creates servers and clients with multiple threads, I prefer to have an explicit setup phase to make it explicit when the copycat scaffolding is created!
I don't like having such complex infrastructure being setup when the singleton is initialized.

vyzo:
Also, having a shutdown (which implies expensive resources) without a setup is kind of incogruous.
If the the context requires a shutdown, then it should also have an explicit setup --- I don't think that expensive resource setup is appropriate for implicit initialization in an object's body.

parkan:
I sort of get where you're coming from but using vars like this is definitely a code smell to me. You also will have a naked NPE if you access `instance` before init

vyzo:
This is just context passed around the test threading. If something ends up being null, the test should fail which should be the case indicating a bug. There is no danger of NPE, the variables are only accessed as opaque references without any data in them.

vyzo:
Also, for the instance pointer triggering an NPE -- this is a local context object to support the test threading. 
There is no chance of being accessed outside the initialization confines to trigger an NPE.

parkan:
Right, so why do you worry about control over init then if it's only used in this limited way? Not gonna argue this further since it's not a very high priority but again, very much against nulls and vars.

PR #38, opened 2016-04-28
very simple sample datastore json example

yusefnapora:
Don't merge me bro!  This is just here to get some eyes on it and make sure I'm basically barking up the right tree, data format wise.

The `picasso.json` file is a simple dummy "datastore", where the refs are just made up strings in a big object e.g. `"Qm001"`.  It has one `Entity` (picasso), and two works, one of which is a derivative of the other; the pen sketch of the starry night is derived from the painting.  Everything is missing signatures, so you can use your imagination there :)

The schema.org stuff is just there so I had some fields to work with for things like `artform`, etc...  I'm not saying we should use that as our data model.  Although... it may actually work alright, given that it was created for search engines, and indexing is our main motivation for extracting / translating raw metadata.  And we do need to decide on some kind of field to indicate what kind of artefact / entity we're talking about.

### DIFF
--- a/sample-data/picasso.json
+++ b/sample-data/picasso.json
@@ -1,0 +1,50 @@
+{
+  "Qm000": { "type": "nil" },
+
+  "Qm001": {
+    "type": "entity",
+    "name": "Pablo Picasso",
+
+    "@context": "http://schema.org/Person",
+    "birthDate": "1881-10-25"
+  },
+
+  "Qm002": {
+    "type": "artefact",
+    "name": "The Starry Night",
+
+    "@context": "http://schema.org/Painting",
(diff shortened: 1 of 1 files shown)


bigs:
lgtm

parkan:
Let's not use the schema.org stuff, that only makes things more confusing

parkan:
what do you think?

yusefnapora:
hmm, yeah I agree that the schema.org stuff kind of muddies the waters here.  I can edit that out.

parkan:
@yusefnapora do you want to update this or just close it for now?

yusefnapora:
oh yeah.  let's just close it, we can revisit later

PR #39, opened 2016-04-28
Bigs cbor

bigs:
in progress, not ready to merge


parkan:
@bigs ok this is abandoned then?

Issue #22, opened 2016-04-26
Copycat state machine


parkan:
Base state machine on top of https://github.com/atomix/copycat

partial implementation in https://github.com/mediachain/mediachain/pull/21/

vyzo:
Merged in the first implementation: The state machine computes the index of the mediachain, tracking the mapping of canonicals to chain heads.

The next part will add the Journal Blockchain generation.

vyzo:
Added Journal Blockchain generation.

vyzo:
The implementation is reasonably complete; remains to be tested in the real world.

vyzo:
initial work is done, closing.

PR #40, opened 2016-04-29
Transactor StateMachine: Journal blockchain generation

vyzo:
#22
- Adds type support for JournalBlocks
- extends the StateMachine to query and generate the journal blockchain every X journal entries

More specifically, the StateMachine API is extended as follows:
- there is a JournalCurrentBlock query that provides access to the current working block
- Every JournalBlockSize entries a new block is generated, stored in the datastore and a reference to the blob is published to clients with a "journal-block" event.

Note that I elected to deterministically generate a block every blocksize entries in order to avoid the pitfalls and complexities of using timers with copycat (it isn't pretty).
The Journal allows the current block to be queried, which is sufficient for a joining client to bootstrap.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
@@ -13,6 +13,8 @@ import cats.data.Xor
 
 object StateMachine {
   import io.mediachain.transactor.Types._
+
+  val JournalBlockSize: Int = 4096 // blocksize for Journal Blockchain
   
   case class JournalInsert(
     record: CanonicalRecord
@@ -34,8 +36,7 @@ object StateMachine {
   case class JournalBlockEvent(ref: Reference) extends JournalEvent
 
   class JournalStateMachine(
-    val datastore: Datastore,
-    val blocksize: Int = 4096
+    val datastore: Datastore
   ) extends CopycatStateMachine with Snapshottable with SessionListener {
     private var seqno: BigInt = 0
     private var index: MMap[Reference, ChainReference] = new MHashMap // canonical -> chain map
@@ -134,7 +135,7 @@ object StateMachine {
     // block generation
     private def blockExtend(entry: JournalEntry) {
       block += entry
-      if (block.length >= blocksize) {
+      if (block.length >= JournalBlockSize) {
         val entries = block.toList
         val newblock = JournalBlock(seqno, blockchain, entries)
         val blockref = datastore.put(newblock)
--- a/transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
@@ -1,7 +1,8 @@
 package io.mediachain.transactor
 
 import scala.collection.mutable.{Set => MSet, HashSet => MHashSet, 
-                                 Map => MMap, HashMap => MHashMap}
+                                 Map => MMap, HashMap => MHashMap,
+                                 ListBuffer}
 
 import io.atomix.copycat.{Command, Query}
 import io.atomix.copycat.server.{Commit, StateMachine => CopycatStateMachine, Snapshottable}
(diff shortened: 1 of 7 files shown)


transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
```diff
@@ -111,6 +124,27 @@ object StateMachine {
...
+        JournalBlock(seqno, blockchain, block.toArray)
+      } finally {
+        commit.release()
+      }
+    }
+
+    // block generation
+    private def blockExtend(entry: JournalEntry) {
+      block += entry
+      if (block.length >= blocksize) {
```
bigs, line 139: this should be exception safe-- put and publish should be transactional

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
```diff
@@ -12,6 +13,8 @@ import cats.data.Xor
 
 object StateMachine {
   import io.mediachain.transactor.Types._
+
+  val JournalBlockSize: Int = 4096 // blocksize for Journal Blockchain
```
bigs, line 17: `JOURNAL_BLOCK_SIZE`

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
```diff
@@ -73,6 +85,7 @@ object StateMachine {
       def commit(ref: Reference, newchain: Reference, oldchain: Option[Reference]) = {
         val entry = ChainEntry(nextSeqno(), ref, newchain, oldchain)
         publishCommit(entry)
+        blockExtend(entry)
```
bigs, line 88: ordering of these might need to be flipped and this should be transactional as weel

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
vyzo, line 139: I still have to release the commit and return the value, so I adopted this pattern to have visual symmetry with the other API entry points.

Alternative is to write it this way:

``` scala
val block = JournalBlock(...)
commit.release()
block
```

which I find uglier (and also incogruous with the other API methods)

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
```diff
@@ -25,15 +28,23 @@ object StateMachine {
...
 
   class JournalStateMachine(
-    val datastore: Datastore
+    val datastore: Datastore,
+    val blocksize: Int = JournalBlockSize // configurable to facilitate testing
   ) extends CopycatStateMachine with Snapshottable with SessionListener {
     private var seqno: BigInt = 0
     private var index: MMap[Reference, ChainReference] = new MHashMap // canonical -> chain map
-    private val clients: MSet[ServerSession] = new MHashSet // this wanted to be called sessions
+    private var block: ListBuffer[JournalEntry] = new ListBuffer      // current block entries
```
bigs, line 44: `mutable.Array` is prob fine

bigs, line 44: or `mutable.Buffer`

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
vyzo, line 17: ugh, I thought Scala constants followed CapitalizedCamelCase pattern.

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
vyzo, line 88: no, the commit needs to be published before the block as it logically happens earlier.

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
vyzo, line 44: I'll change to mutable.Buffer, seems like a better class for the purpose (I didn't know it existed :)

parkan, line 44: ListBuffer is fine, we don't need random access right?

bigs:
@vyzo the change itself looks good. i think, however, that our tests should not involve copycat at all. the logic associated w/ our statemachine can be clearly separated out from the networking/consensus bits. we can rely on copycat's tests for that.

furthermore, i'm thinking that we should move the logic layer to be completely stateless/immutable. i have a pretty good example of how this works, but the general idea is this:
- separate all state into a container `case class State` or something of the like. this includes buffer, etc
- all state machine logic functions accept **the current state in addition to any arguments from the client** and return the **updated state along with a list of any effectful actions we need to take**
- the "driver" then persists the state to a `var` and iterates through all of the "actions", executing them (i.e. update all sessions w/ this update, etc)

this separation allows us to completely unit test our functionality and makes it MUCH simpler to ensure transactionality (i.e. don't commit to the serialized `var` until we have successfully written entries to the datastore, etc)

i feel pretty strongly that we should make a change like this earlier rather than later. it will make our logic much simpler to interact with

bigs:
i have an example of logic like this that i can share w/ u if you want something to work off of. can do a call too

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
vyzo, line 44: no, just a append and toArray operations.

parkan, line 44: Oh, if we need to convert to a java array anyway we should use an ArrayBuffer

vyzo:
Uhm, the state immutability idea is impractical. Copycat is designed to provide a replicated state machine abstraction, and state machines are stateful by definition.

I agree that we need tests that only test the logic of the journal state machine, but on the same time we also need tests to specify and test the copycat implementation!

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
vyzo, line 44: ok, keeping ListBuffer -- it does the job just fine after all.

PR #45, opened 2016-05-02
CBOR serialization with cbor-java

yusefnapora:
Serializes DataObjects and JournalEntries to/from a `CValue` type that wraps the `DataItem`s used by the [cbor-java](https://github.com/c-rack/cbor-java) library.  See #36, which was opened against a different branch.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -11,7 +11,7 @@ object MediachainBuild extends Build {
     scalacOptions ++= Seq("-Xlint", "-deprecation", "-Xfatal-warnings", "-feature"),
     libraryDependencies ++= Seq(
       "org.typelevel" %% "cats" % "0.4.1",
-      "org.json4s" %% "json4s-jackson" % "3.2.11",
+      "org.json4s" %% "json4s-jackson" % "3.3.0",
       "org.specs2" %% "specs2-core" % "3.7" % "test",
       "org.specs2" %% "specs2-junit" % "3.7" % "test",
       "org.specs2" %% "specs2-matcher-extra" % "3.7" % "test",
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -4,6 +4,9 @@ import Keys._
 object MediachainBuild extends Build {
   updateOptions := updateOptions.value.withCachedResolution(true)
 
+  val specs2Version = "3.7.3"
+  val scalaCheckVersion = "1.13.1"
+
   override lazy val settings = super.settings ++ Seq(
     organization := "io.mediachain",
     version := "0.0.1",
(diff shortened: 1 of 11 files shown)


project/Build.scala
```diff
@@ -4,18 +4,22 @@ import Keys._
...
     scalaVersion := "2.11.7",
     scalacOptions ++= Seq("-Xlint", "-deprecation", "-Xfatal-warnings", "-feature"),
     libraryDependencies ++= Seq(
       "org.typelevel" %% "cats" % "0.4.1",
-      "org.json4s" %% "json4s-jackson" % "3.2.11",
-      "org.specs2" %% "specs2-core" % "3.7" % "test",
-      "org.specs2" %% "specs2-junit" % "3.7" % "test",
-      "org.specs2" %% "specs2-matcher-extra" % "3.7" % "test",
-      "org.scalacheck" %% "scalacheck" % "1.13.0" % "test"
+      "org.json4s" %% "json4s-jackson" % "3.3.0",
```
vyzo: do we still want the json4s dependency?

yusefnapora: ahh, good point.  if we remove the conversions from JValue to CValue we can remove it.  I don't think we particularly need the conversions...

parkan: dumb question: is there an easy way to dump out cbor as something human-readable for debugging?

transactor/src/main/scala/io/mediachain/transactor/TypeSerialization.scala
```diff
@@ -0,0 +1,243 @@
...
+  def referenceFromCMap(cMap: CMap): Xor[DeserializationError, Reference] = {
+    val linkOpt: Option[CValue] =
+      cMap.asStringKeyedMap.get("@link")
+
+    Xor.fromOption(linkOpt, ReferenceDecodingFailed("@link field not found"))
+      .flatMap(decodeLinkValue)
+  }
+
+
+  def decodeLinkValue(linkVal: CValue): Xor[DeserializationError, Reference] =
```
parkan, line 234: @yusefnapora I'm not following what's going on here. Where does `dummy@$seqno` come from?

vyzo, line 234: that's the toString representation of dummy references

parkan, line 234: ok, this makes sense now, but I'd really prefer not to have these type conversions depend on Dummies. Any thoughts?

yusefnapora, line 234: Yeah, I don't like this either....  is it possible to define a static method inside a trait?  Seems like probably not, since it would really be a method on the companion object... but it would be nice to have a `fromCValue` method on the `CborSerializable` trait that does the actual unpacking.

parkan, line 234: Yeah, companion objects are how you implement static methods on traits.

yusefnapora, line 234: Then maybe we could define a `CborDeserializer[T <: CborSerializable]` trait?  And have companion objects for each type implement it.  That might be better than the "giant switch statement" I'm using here...

vyzo, line 234: it would also be nice to have a hash table dispatch instead of the giant switch statement

transactor/src/main/scala/io/mediachain/transactor/TypeSerialization.scala
```diff
@@ -0,0 +1,243 @@
...
+    val typeNameOpt: Option[String] =
+      cMap.getAs[CString]("type").map(_.string)
+
+    Xor.fromOption(typeNameOpt, TypeNameNotFound())
+      .flatMap(name => fromCMap(cMap, name))
+  }
+
+
+  def fromCMap(cMap: CMap, typeName: String)
+  : Xor[DeserializationError, CborSerializable] = typeName match {
```
parkan, line 86: so you can write `(typeName: @switch)` here to make sure it's optimized

yusefnapora, line 86: neato 😄

PR #47, opened 2016-05-02
Spec for CBOR conversion of transactor types

yusefnapora:
This just checks that the transactor types get serialized with the expected type names, and that a round-trip conversion results in equal objects.

Figured I should do this before refactoring the deserialization code :)


parkan:
lgtm

will need to go over how some of these types are used tomorrow, particularly Entity and Artefact (I thought they were mostly used in commands but I guess they're also used by the index?)

otherwise I think we have a very nice way of delineating interface/implementation/serialization right around the corner

Issue #30, opened 2016-04-27
Chain Persistence


bigs:
- [x] In memory
- [x] RocksDB
- [x] DynamoDB
- [ ] IPFS

parkan:
closing pending IPFS work

PR #48, opened 2016-05-02
Datastorage impl

bigs:
Related to #30. So far, RocksDB is the only impl.

Clean up before merge:
- [x] Remove redundant multihash code


transactor/src/main/scala/io/mediachain/transactor/TypeSerialization.scala
```diff
@@ -41,7 +41,7 @@ object TypeSerialization {
 
   def dataObjectFromCbor(cValue: CValue): Xor[DeserializationError, DataObject] =
     fromCbor(cValue) match {
-      case Xor.Left(err) => Xor.left(err)
+      case l@Xor.Left(_) => l
```
vyzo: what does this do?

transactor/src/main/scala/io/mediachain/util/datastore/RocksDatastore.scala
```diff
@@ -0,0 +1,37 @@
...
+    ref
+  }
+
+  override def close: Unit = {
+    db.close()
+  }
+}
+
+object RocksDatastore {
+  class RocksReference(val id: Array[Byte]) extends Reference {
```
vyzo: You need equals and hashCode for the references, they are used as hash keys.

vyzo: Also, it would be nice to have a toString that prints the base58 or multihash of the array for debugging.

transactor/src/main/scala/io/mediachain/util/datastore/RocksDatastore.scala
```diff
@@ -0,0 +1,37 @@
...
+import io.mediachain.transactor.Types.{DataObject, Datastore, Reference}
+import io.mediachain.util.cbor.CborAST.{CString, CValue}
+import io.mediachain.hashing.MultiHash
+
+class RocksDatastore(path: String) extends Datastore with AutoCloseable {
+  import RocksDatastore.RocksReference
+
+  val db = RocksDB.open(path)
+
+  override def put(obj: DataObject): Reference = {
```
vyzo: can we also get a get method for debugging purposes?

transactor/src/main/scala/io/mediachain/transactor/TypeSerialization.scala
parkan: binds the whole match to l, so `l = Xor.Left(thing)` in this case

transactor/src/main/scala/io/mediachain/util/datastore/RocksDatastore.scala
```diff
@@ -29,6 +29,13 @@ object RocksDatastore {
   class RocksReference(val id: Array[Byte]) extends Reference {
     override val CBORType: String = ""
     override def toCbor: CValue = CString(new String(id))
+
+    override def toString(): String =
+      MultiHash.fromBytes(id)
+        .map(_.toString)
+        .getOrElse("RocksReference[Invalid reference]")
+
+    override def hashCode(): Int = id.hashCode
```
vyzo: uhm, we need equals too

transactor/src/main/scala/io/mediachain/util/datastore/RocksDatastore.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  override def close: Unit = {
+    db.close()
+  }
+}
+
+object RocksDatastore {
+  class RocksReference(val id: Array[Byte]) extends Reference {
+    override val CBORType: String = ""
```
vyzo: why is this an empty CBORType? I take it we are doing pojo if we need to serialize the Rocks?

bigs: we override the serialization to just create a CString

transactor/src/main/scala/io/mediachain/util/datastore/RocksDatastore.scala
bigs: done

transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
```diff
@@ -17,19 +19,20 @@ object Dummies {
...
+    override def put(obj: DataObject): MultiHash = {
+      val ref = MultiHash.hashWithSHA256(seqno.toString.getBytes)
       seqno += 1
       store += (ref -> obj)
       ref
     }
-    
-    def get(ref: Reference) = store.get(ref)
+
+    def get(ref: MultiHash): Option[DataObject] = store.get(ref)
```
vyzo: what is this? are you dropping the reference type?

transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
```diff
@@ -17,19 +19,20 @@ object Dummies {
...
   }
-  
+
   class DummyStore extends Datastore {
     var seqno = 0
-    val store: MMap[Reference, DataObject] = new MHashMap
-    
-    override def put(obj: DataObject): Reference = {
-      val ref = new DummyReference(seqno)
+    val store: MMap[MultiHash, DataObject] = new MHashMap
```
vyzo: and here, why this? We still need Reference -> Datastore mapping in order to be able to extract values from the datastore for testing.

transactor/src/main/scala/io/mediachain/transactor/Types.scala
```diff
@@ -213,6 +215,15 @@ object Types {
   
   // Datastore interface
   trait Datastore {
-    def put(obj: DataObject): Reference
+    def put(obj: Array[Byte], hash: MultiHash): Unit
+
+    def put(obj: DataObject): MultiHash = {
```
vyzo: if you change the return type to Multihash you effectively drop References as a type and they are used in quite a lot of places.
Let this return references and encapsulate the multihash.

vyzo:
@bigs I don't like you dropping the Reference type from the Datastore interface -- please keep it and encapsulate the multihash inside the references.
Also, in the Datastore interface I would prefer to keep the simplified interface and have the binary put in a subinterface:

``` scala
trait BinaryDatastore extends Datastore {
  put(DataObject) : Refence = {...}
  put(bytes) : Reference
}

```

bigs:
A note for @yusefnapora, @parkan, @vyzo 

This should be ready, pending a merge of @yusefnapora's branch re: chain types. IMO it needs a little cleaning (looks great, but I think we could cut out some redundant implicits?) before going in, so I didn't pull the trigger myself.

With that merged in, implementing the `get` methods for `RocksDatastore` and `DynamoDatastore` should be no problem.

vyzo:
The path dependent type breaks usage for Datastore.get -- the branch can't even compile the tests as is right now.

vyzo:
superseded by #56

PR #50, opened 2016-05-03
Copycat cluster test spec

vyzo:
Adds a test spec for a cluster of copycat servers and disables JournalBlockchainSpec2.

The new test creates a cluster with 3 servers and tests with 3 clients each connected to a different server.

### DIFF
--- a/transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
+++ b/transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
@@ -50,27 +50,25 @@ object DummyClusterContext {
   def setup(address1: String, address2: String, address3: String,
             blocksize: Int = StateMachine.JournalBlockSize) = {
     println("*** SETUP DUMMY COPYCAT CLUSTER")
-    val store1 = new Dummies.DummyStore
-    val logdir1 = DummyContext.setupLogdir()
-    val server1 = Copycat.Server.build(address1, logdir1, store1, blocksize)
-    server1.bootstrap().join()
-    val store2 = new Dummies.DummyStore
-    val logdir2 = DummyContext.setupLogdir()
-    val server2 = Copycat.Server.build(address2, logdir2, store2, blocksize)
-    server2.join(new Address(address1)).join()
-    val store3 = new Dummies.DummyStore
-    val logdir3 = DummyContext.setupLogdir()
-    val server3 = Copycat.Server.build(address3, logdir3, store3, blocksize)
-    server3.join(new Address(address1), new Address(address2)).join()
-    val client1 = Copycat.Client.build()
-    client1.connect(address1)
-    val client2 = Copycat.Client.build()
-    client2.connect(address2)
-    val client3 = Copycat.Client.build()
-    client3.connect(address3)
-    DummyClusterContext(Array(DummyContext(server1, client1, store1, logdir1),
-                              DummyContext(server2, client2, store2, logdir2),
-                              DummyContext(server3, client3, store3, logdir3)))
+    val dummies = Array(address1, address2, address3)
+      .map { address =>
+        val store = new Dummies.DummyStore
+        val logdir = DummyContext.setupLogdir()
+        val server = Copycat.Server.build(address, logdir, store, blocksize)
+        val client = Copycat.Client.build()
+        DummyContext(server, client, store, logdir)
+    }
+    
+    // start cluster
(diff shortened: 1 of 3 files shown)


transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
```diff
@@ -40,3 +41,41 @@ object DummyContext {
...
+        val store = new Dummies.DummyStore
+        val logdir = DummyContext.setupLogdir()
+        val server = Copycat.Server.build(address, logdir, store, blocksize)
+        val client = Copycat.Client.build()
+        DummyContext(server, client, store, logdir)
+    }
+    
+    // start cluster
+    dummies(0).server.bootstrap().join()
+    dummies(1).server.join(new Address(address1)).join()
```
parkan, line 64: it's not at all confusing that these are two completely different `join`s, thanks java 😓

transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
```diff
@@ -40,3 +41,41 @@ object DummyContext {
...
+object DummyClusterContext {
+  def setup(address1: String, address2: String, address3: String,
+            blocksize: Int = StateMachine.JournalBlockSize) = {
+    println("*** SETUP DUMMY COPYCAT CLUSTER")
+    val dummies = Array(address1, address2, address3)
+      .map { address =>
+        val store = new Dummies.DummyStore
+        val logdir = DummyContext.setupLogdir()
+        val server = Copycat.Server.build(address, logdir, store, blocksize)
+        val client = Copycat.Client.build()
```
parkan, line 58: am given that the client doesn't seem to take any contextual params, does it make sense for it to be in the context?

what address is used for outgoing connection?

vyzo, line 58: the client doesn't connect until you call connect, which happens after the dummy construction (later in the setup function)

transactor/src/test/scala/io/mediachain/transactor/JournalClusterSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+  }
+  
+  def afterAll() {
+    JournalClusterSpecContext.shutdown()
+  }
+  
+  def insertAndUpdateJournal = {
+    val context = JournalClusterSpecContext.context
+    val ops = context.cluster.dummies.map(insertAndUpdate(_))
+    val res = ops.map(Await.result(_, timeout))
```
parkan, line 42: We should probably use an XorF monad transformer for this in production (though it's probably fine here, since the timer blowing will be caught by the test harness)

vyzo, line 42: yeah, a timer exception is intended to fail the test.

transactor/src/test/scala/io/mediachain/transactor/JournalClusterSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+      case Xor.Right(tuple) => tuple
+    }
+    
+    (res(0) must beRightXor) and
+    (res(1) must beRightXor) and
+    (res(2) must beRightXor)
+  }
+  
+  private def insertAndUpdate(dummy: DummyContext)
+  : Future[Xor[JournalError, (Reference, Reference)]] = {
```
parkan, line 54: Yeah, we should replace these `Future[Xor...` with a `XorF` to make the nesting less insane http://eed3si9n.com/herding-cats/stacking-future-and-either.html#XorT+datatype

transactor/src/test/scala/io/mediachain/transactor/JournalClusterSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+    dummy.client.insert(Entity(Map()))
+      .flatMap { 
+        case err: Xor.Left[JournalError] => Future { err }
```
parkan, line 57: I started rewriting this but ultimately we should juse use XorF

transactor/src/test/scala/io/mediachain/transactor/DummyContext.scala
parkan, line 58: right, I meant what the client addr/port

parkan:
This is good overall, we can clean up a lot of logic with transformers down the line

PR #51, opened 2016-05-03
cbor deserialization refactor

yusefnapora:
- Adds a `CborDeserializer[T]` trait that's extended by singleton objects for each `CborSerializable` type we want to decode.  
- Replaces the `typeName match { ... }`  construct in the `fromCMap` method with a `Map[String, CborDeserializer[CborSerializable]]` which is used to look up the deserializer for a given type name.
- Removes the `CborSerializable` trait from the `ChainReference` types, since they're internal to the transactor.
- Renames the `TypeSerialization` object to `CborSerialization`, since we also have pojo serialization for some types.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
@@ -14,10 +14,11 @@ object CborSerialization {
 
   import scala.util.Try
 
-  type DeserializerMap = Map[String, CborDeserializer[CborSerializable]]
+  type DeserializerMap = Map[CborTypeName, CborDeserializer[CborSerializable]]
 
   /**
     * Try to deserialize a `DataObject` from a cbor `CValue`
+ *
     * @param cValue the `CValue` to decode
     * @return a `DataObject`, or `DeserializationError` on failure
     */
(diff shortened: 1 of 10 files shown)


transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
```diff
@@ -0,0 +1,452 @@
...
+    */
+  def fromCMap(cMap: CMap)
+  : Xor[DeserializationError, CborSerializable] = {
+
+    // TODO: create multiple deserializer maps for different contexts
+    // e.g. DataStore should deserialize chain cells to specific subtypes, etc
+    val transactorDeserializers
+    : Map[String, CborDeserializer[CborSerializable]] = Map(
+      CBORTypeNames.Entity -> EntityDeserializer
+        .asInstanceOf[CborDeserializer[CborSerializable]],
```
parkan: what happens if you don't `.asInstanceOf`, is the type hint on the map not enough?

I might do something like

``` scala
Seq(
  CBORTypeNames.Entity -> EntityDeserializer,
  CBORTypeNames.Artefact -> ArtefactDeserializer,
  ...
).map(t => (t._1, t_2.asInstanceOf[CborDeserializer[CborSerializable]]).toMap
```

but that's just me 😄

parkan: oh and I would also construct this map at the toplevel of the object, not inside the method def

yusefnapora: ah, that Seq / map trick looks nice :)

good call about moving to the top level

yusefnapora: I get an odd type error without the cast, btw... 

```
[error]  found   : scala.collection.immutable.Map[String,io.mediachain.transactor.CborSerialization.CborDeserializer[_ >: io.mediachain.transactor.Types.JournalBlock with io.mediachain.transactor.Types.ChainEntry with io.mediachain.transactor.Types.CanonicalEntry with io.mediachain.transactor.Types.ArtefactChainCell with io.mediachain.transactor.Types.EntityChainCell with io.mediachain.transactor.Types.Artefact with io.mediachain.transactor.Types.Entity <: Product with Serializable with io.mediachain.transactor.CborSerialization.CborSerializable]]
[error]  required: Map[String,io.mediachain.transactor.CborSerialization.CborDeserializer[io.mediachain.transactor.CborSerialization.CborSerializable]]
```

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
```diff
@@ -0,0 +1,452 @@
...
+    } yield value
+  }
+
+
+  /**
+    * String constants for the `type` field used to indicate the type of
+    * object represented by a cbor map.
+    */
+  object CBORTypeNames {
+    val Entity = "entity"
```
parkan: I like this pattern overall but something about using strings like this bothers me. I feel like we should use symbols, if we're going the more dynamic route, or case objects for more scala-style. Case objects serialize/toString to their names and you get exhaustiveness checking so I'm kind of leaning towards them. This works fine (and I suspect is faster internally because it's an address comparison):

``` scala
scala> object Foo {
     | case object Bar
     | case object Baz
     | }
defined object Foo

scala> val m = Map(Foo.Bar -> "123", Foo.Baz -> "456")
m: scala.collection.immutable.Map[Product with Serializable,String] = Map(Bar -> 123, Baz -> 456)

scala> m.get(Foo.Baz)
res4: Option[String] = Some(456)

scala> Foo.Baz.hashCode
res5: Int = 66555
```

yusefnapora: oh, that's cool.  I didn't realize you could use case objects like that.  That seems much better than using strings to me.

parkan: that's pretty much what case objects are for :)

you can also give them a supertype if you want something nicer than Product

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
```diff
@@ -91,55 +92,77 @@ object CborSerialization {
...
-      * The set of all valid ArtefactChainCell type names (including subtypes)
-      */
-    val ArtefactChainCellTypes: Set[String] = Set(
+  sealed trait CborTypeName {
+    val stringValue: String
+    def cborString: CString = CString(stringValue)
+    override def toString = stringValue
+  }
+
+  object CborTypeNames {
```
vyzo: perhaps call it CborType instead of CborTypeName?

yusefnapora:
@bigs - the last commit is probably relevant to the datastore work you're doing; I made the `CBORType` value in `CborSerializable` optional, and it's now represented with case objects that extend the `CborTypeName` trait.

Let me know if you think there's a better way to do this btw :)

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
```diff
@@ -341,19 +362,20 @@ object CborSerialization {
...
+ *
     * @param cMap a cbor `CMap` to check the type of
     * @param typeName the required value of the `type` field
     * @return `Unit` on success, or `DeserializationError` if there is no
     *        `type` field, or if the value is incorrect
     */
-  def assertRequiredTypeName(cMap: CMap, typeName: String)
+  def assertRequiredTypeName(cMap: CMap, typeName: CborTypeName)
   : Xor[DeserializationError, Unit] = {
     getTypeName(cMap)
```
parkan:
``` scala
if(getTypeName(cMap).exists(_ == typeName)){
...
}
```

bigs:
@yusefnapora this is perfect, thanks!

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
yusefnapora: oh yeah :)  for some reason I always forget that `.exists` exists 😆

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
```diff
@@ -367,32 +389,35 @@ object CborSerialization {
     *        `type` field, or if the value is not contained in the
     *        `typeNames` set
     */
-  def assertOneOfRequiredTypeNames(cMap: CMap, typeNames: Set[String])
+  def assertOneOfRequiredTypeNames(cMap: CMap, typeNames: Set[CborTypeName])
   : Xor[DeserializationError, Unit] = {
     getTypeName(cMap)
```
parkan: same here

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
```diff
@@ -0,0 +1,479 @@
...
+ *
+    * @param cMap a cbor `CMap` to get the type name from
+    * @return the value of the `type` field, or a `DeserializationError` error if
+    *         no `type` field exists, or its value is not a valid type name
+    */
+  def getTypeName(cMap: CMap): Xor[DeserializationError, CborTypeName] =
+    getRequired[CString](cMap, "type")
+      .map(_.string)
+      .leftMap(_ => TypeNameNotFound())
+      .flatMap(CborTypeNames.fromString)
```
parkan:
``` scala
.bimap(err => TypeNameNotFound(), res => res.string)...
```

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
```diff
@@ -0,0 +1,479 @@
...
+    * @param cMap the cbor `CMap` to decode as a `Reference`
+    * @return the decoded `Reference`, or a `DeserializationError` if decoding
+    *         fails
+    */
+  def referenceFromCMap(cMap: CMap): Xor[DeserializationError, Reference] =
+    getRequired[CMap](cMap, "@link")
+      .flatMap { linkVal: CValue =>
+        linkVal match {
+          case CString(s) if s.startsWith("dummy@") => {
+            DummyReferenceDeserializer.fromCMap(cMap)
```
parkan: @bigs what was the conclusion you reached regarding reference types? I still want to get this DummyReference out of here

yusefnapora: Do we really need separate `Reference` subtypes?  Maybe we can just make the `DummyStore` and the rocksdb store use the same type as the eventual IPLD store, and encode everything to an IPLD-style `@link` map...  We can add a catch-all `meta` map for link-properties, in case a store implementation needs to store some arbitrary value besides the actual reference.

parkan: DummyRef specialization gives us the nice sequential properties but we don't seem to really use this anywhere. I definitely envisioned storage as a black box with a generic reference type.

yusefnapora: Well.... we could still use an integer sequence number by just stuffing it into the `meta` map and ignore the actual value of the `@link` field... but that would be hidden in the `DummyStore` implementation.

Or we could just use multihashes for the dummy store also, which I kind of prefer, since it's closer to the real implementation.

parkan: Yeah tbh I was thinking multihash for everything

vyzo: multihash is fine, but we shouldn't have meta in the references -- they should be a naked {"@link":...} ultimately.

transactor/src/main/scala/io/mediachain/transactor/CborSerialization.scala
yusefnapora: @vyzo, it occurred to me this morning that MediachainType might be a better name...  CborType makes me think of e.g. CInt, CString, etc (the various cbor primitives).   What do you think?

vyzo: yeah, that might be a better option.

vyzo:
:+1: for multihash references.

PR #52, opened 2016-05-03
Use XorT to dramatically clean up nested Future/Xor ops

parkan:
power overwhelming

using the Power of Category Theory™, we can make the syntax around processing nested Xors/Futures (&c) much nicer. I've applied only a sprinkling of it here since most other callsites currently just await on the result, but I also have as stash where I have the underlying methods XorT'ed and other callsites adapted with `.value` if you think we should do this throughout.


vyzo:
:+1: 
Power Overwhelming.

PR #54, opened 2016-05-04
add subtypes for chain cells

yusefnapora:
putting this up early to get some feedback.

I split the transactor and datastore types into separate objects in the new `io.mediachain.types` package.   This should make it easier to pull into a separate project later.  `DatastoreTypes` has everything that inherits from `DataObject`, and `TransactorTypes` has the transactor-internal types like `ChainReference` and the journal interfaces.  The `DatastoreTypes` are all `CborSerializable`, but the `TransactorTypes` use pojo serialization.

Also changes the `EntityChainCell` and `ArtefactChainCell` to regular classes and adds some case classes that extend them (only a few so far).

the existing tests pass, but I haven't added specs for unpacking the chain cell subtypes yet.


transactor/src/main/scala/io/mediachain/types/Datastore.scala
```diff
@@ -0,0 +1,8 @@
+package io.mediachain.types
+
+// Datastore interface
+trait Datastore {
+  import DatastoreTypes.{DataObject, Reference}
+
+  def put(obj: DataObject): Reference
+}
```
vyzo: no newline at end of file

transactor/src/main/scala/io/mediachain/types/DatastoreTypes.scala
```diff
@@ -0,0 +1,202 @@
+package io.mediachain.types
+
+
+object DatastoreTypes {
```
vyzo: can't we call this just Datastore and put the Datastore interface together as well?
DatastoreTypes is a very ugly name, and it is also rather sad to have a forlorn Datastore trait with one method in its own file.

transactor/src/main/scala/io/mediachain/types/TransactorTypes.scala
```diff
@@ -0,0 +1,54 @@
+package io.mediachain.types
+
+import cats.data.Xor
+
+import scala.concurrent.Future
+
+
+object TransactorTypes {
```
vyzo: similarly, let's call this just Transactor.

transactor/src/main/scala/io/mediachain/types/DatastoreTypes.scala
yusefnapora: sounds good to me :)

vyzo: it is a member of a package named io.mediachain.types after all, so types.Datastore and types.Transactor is natural.

vyzo:
I like the idea of putting the chain cell types in the meta as a field, it's a flexible approach.

transactor/src/main/scala/io/mediachain/types/CborSerialization.scala
```diff
@@ -1,51 +1,123 @@
...
+    * where you call the various `fromCbor` methods.
+    *
+    * e.g:
+    * ```
+    * implicit val deserializerMap = CborSerialization.transactorDeserializers
+    *
+    * val fooCell = fromCborBytes(fooCellBytes)
+    * ```
+    */
   type DeserializerMap = Map[MediachainType, CborDeserializer[CborSerializable]]
```
bigs, line 35: what if we took the witness approach to this... this way the passing around of `deserializers` wouldn't be so shifty

``` scala
  trait Deserializer[T] {
    def deserialize(bytes: Array[Byte]): Option[T]
  }

  // dumb string deserialization
  implicit val deserializeInt = new Deserializer[Int] {
    override def deserialize(bytes: Array[Byte]): Option[Int] = {
      val str = new String(bytes)
      try {
        str.toInt
      } catch {
        case _ => None
      }
    }
  }

  def foo(myInt: Array[Byte])(implicit d: Deserializer[Int]): Int = {
    d.deserialize(myInt).map(_ + 1).getOrElse(0)
  }

  def generalDeserializer[T](implicit d: Deserializer[T]): Option[T] = {
     val bytes = fetchDb()
     d.deserialize(bytes)
  }
```

bigs, line 35: ah realizing this is for reifying types

bigs, line 35: ok so this impl is actually fine but i'm not sure that u have to pass around the implicit `derserializers` everywhere. it's really confusing to see it not explicitly passed in to all the fns that accept it as a parameter

vyzo:
For the new chain cell some classes, we will need to add a cons method that constructs the exact type for use inside the state machine.
I will add it and send a PR on top of this.

vyzo:
PR #55 for supporting chain cell subtypes in the state machine.

transactor/src/main/scala/io/mediachain/transactor/Dummies.scala
```diff
@@ -3,7 +3,8 @@ package io.mediachain.transactor
 import scala.collection.mutable.{HashMap => MHashMap, Map => MMap}
 
 object Dummies {
-  import io.mediachain.transactor.Types._
+  import io.mediachain.types.Datastore
```
vyzo: this is redundant

yusefnapora: ah, good catch, thanks

vyzo:
:+1:

transactor/src/main/scala/io/mediachain/transactor/StateMachine.scala
```diff
@@ -97,15 +98,15 @@ object StateMachine {
           (cref, cell) match {
             case (EntityChainReference(chain), EntityChainCell(entity, xchain, meta)) => {
               if (checkUpdate(ref, chain, entity, xchain)) {
-                val newcell = EntityChainCell(ref, chain, meta)
+                val newcell = cell.cons(chain)
```
parkan, line 101: in scala `cons` is called `::`

transactor/src/main/scala/io/mediachain/types/CborSerialization.scala
parkan, line 35: pretty standard, same as the json4s type conversions etc

transactor/src/test/scala/io/mediachain/transactor/CborSerializationSpec.scala
```diff
@@ -43,12 +55,59 @@ object CborSerializationSpec extends BaseSpec {
       meta = Map("created" -> CString("the past"))
     )
 
+    val entityUpdateCell = EntityUpdateCell(
+      entity = multihashRef("foo"),
+      chain = Some(multihashRef("bar")),
+      meta = Map("favoriteColor" -> CString("Blue! No, Yellow!!!"))
```
parkan, line 61: it's really fun to come up with these but I would stick them into a list and pull one out at random for each test, similar to how we do it in the L-SPACE fixtures https://github.com/mediachain/L-SPACE/blob/master/l_space/src/test/scala/io/mediachain/GraphFixture.scala#L47

I would also inject some unicode in here

yusefnapora, line 61: that's a good call.  Do you think it's worth using scalacheck to generate these?

parkan: I looked into scalacheck generators but they don't seem to be well suited for this task. I think the ad-hoc bags of words are OK for now; we can swap them out for being backed by more sophisticated generators later on.

PR #55, opened 2016-05-05
yn-chain-types: support chain cell subtypes in state machine

vyzo:
#54
- adds a cons method to ChainCells for consing on top of a chain
- modifies the state machine to use cons for constructing the actual chain cells


yusefnapora:
awesome, I like it :)

PR #56, opened 2016-05-05
basic datastore implementation

vyzo:
Merged in the tip of cbor chain types and deserialization development, and beat to shape bigs' initial implementation.

Notes and Issues:
The dynamodb implementation is very naive as it blocks the current thread and can result in runtime errors from the dynamodb call.
Also, it doesn't take into account dynamo limits in object size, which would require us to chunk the objects.


vyzo:
branch is all screwed up with the monorepo merge.
Tombstoning and hard resetting a new branch to pull the datastore implementation cleanly into master.

PR #58, opened 2016-05-05
WIP: Move types to own project

parkan:


project/Build.scala
```diff
@@ -32,12 +32,28 @@ object MediachainBuild extends Build {
   ))
 
   lazy val transactor = Project("transactor", file("transactor"))
-    .settings(settings)
-    .dependsOn(scalaMultihash)
+    .settings(settings ++ Seq(
+      libraryDependencies ++= Seq(
+        "io.atomix.copycat" % "copycat-server" % "1.0.0-rc7",
```
vyzo: do we really want copycat as a dependency everywhere?
I think it might be better to isolate it in the transactor built.sbt

PR #60, opened 2016-05-05
Monorepo

parkan:
After exploring 3 (4?) different ways of pulling in everything as submodules, subprojects, subtrees, nested sbt projects etc I've decided that it's just not worth it, so here's a monorepo with a single layer of sbt projects. Build.scala is very messy right now, but everything should compile and tests pass.

Downsides:
- big repo is big
- because circle works on a per-repo basis (more or less), we end up with slower CI -- I think I have a way around this, though
- mediachain-cli no longer shares submodule protobuf definitions

I think I can fix two out of these three problems by setting up some post-receive hooks, stay tuned.


parkan:
Oh, I guess another downside is that commits will no longer be scoped to a particular slack channel :(

But this may be fixable the same way, possibly

vyzo:
@parkan  can you avoid doing any refactoring to transactor/types stuffs while merging this monster?
Just limit to monorepo and changes to Build.scala etc, but no other tip changes or it will be very hard to follow what's going on here.

Issue #61, opened 2016-05-06
DynamoDatastore implementation Refinement


vyzo:
The initial implementation of the DynamoDatastore in #56 needs to be significantly refined to be production-ready.

In order to support system operation and deal with failures/throughput constraints and deterministic semantics for the state machine, we need a few things:
- Implement queued writes in the background, with local storage providing buffer while writes to dynamo are in progress to reduce memory pressure and recover from transient API failures
- Implement batched read/writes for chunked data. The current implementation performs separate requires, which is inefficient. Alert: the batch dynamo API is a real PITA to work with in Scala.
- Ensure writes are idempotent and persistent with operational support for failure recovery.

vyzo:
#68 implements dynamo write-back with local cache in rocksdb.

vyzo:
Also, experimenting determined that dynamo writes are idempotent as expected.

vyzo:
Potential refinement for PersistentDatastore is to use multiple writer threads.

vyzo:
Multiple writer threads implemented in #80 
Closing.

PR #62, opened 2016-05-06
DynamoDatastore: batch reads

vyzo:
very ugly compared to the unbatched code, but potentially a lot faster.

One point of possible refinement is the busy loop when dynamo returns a partial response in getChunksBatch.
This could use an exponential back-off or be futurized.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
+++ b/transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
@@ -2,9 +2,11 @@ package io.mediachain.datastore
 
 import com.amazonaws.auth.BasicAWSCredentials
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient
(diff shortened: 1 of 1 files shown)


yusefnapora:
I was going to say that partial responses seem unlikely at the sizes we're working with, but I suppose they could happen if many clients were trying to access the chunks table concurrently & we hit a provisioning limit....  

 Exponential backoff is probably a decent idea, but that seems like a job for a separate PR that futurizes the blocking calls.

As unfortunate as the batch API is, I'm cool with using it if it works.  For reads anyway; if the batch writes are too complex maybe we should just avoid them for now

vyzo:
the batch writes may be paritally successful, returning an ugly jumble indicating which items weren't written.
This requires failure recovery, with some exponential back off etc -- and this is better suited to a background thread doing queued writes.
I made an issue -  #61

vyzo:
Closing and parking this for now.

PR #65, opened 2016-05-09
basic datastore implementation [reset]

vyzo:
This is a reset branch prior to monorepo merging mess up in #56 
Provides a basic implementation for datastores with Rocks and DynamoDb


transactor/src/main/scala/io/mediachain/datastore/BinaryDatastore.scala
```diff
@@ -0,0 +1,39 @@
...
+        case Xor.Right(obj) => obj
+        case Xor.Left(err) => {
+          // this is probably indicative of a bug, so throw
+          throw new DatastoreException("Object deserialization error: " + err.message)
+        }
+      }
+    }
+  }
+  
+  override def put(obj: DataObject): Reference = {
```
parkan, line 30: I'm down with this being polymorphic but the return types should prob still be the same, unless the other `put` is meant to be `private`?

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
```diff
@@ -0,0 +1,158 @@
...
+
+// TODO error handling
+//      queued eventual writes in the background with disk backing of 
+//       in-progress writes
+//      use batch writes for chunked puts
+class DynamoDatastore(table: String, creds: BasicAWSCredentials)
+  extends BinaryDatastore with AutoCloseable {
+  import scala.collection.JavaConversions._
+  
+  val chunkSize = 1024 * 384 // 384 KB; DynamoDB has 400KB limit
```
parkan, line 19: I really like https://github.com/twitter/util#space though I am guessing you will scoff at it 😛

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
```diff
@@ -0,0 +1,158 @@
...
+  val chunkTable = table + "Chunks"
+  
+  val db = new AmazonDynamoDBClient(creds)
+
+  override def put(key: MultiHash, value: Array[Byte]) {
```
parkan: so this is the `Unit` put... right?

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
```diff
@@ -0,0 +1,158 @@
...
+    db.putItem(table, item)
+  }
+
+  private def putChunked(key: MultiHash, value: Array[Byte]) {
+    val key58 = key.base58
+    val chunks = chunkCount(value.length)
+    val chunkIds = (1 to chunks).map { n => key58 + "#" + n }
+    
+    // this should use the BatchWriteItemRequest API (PITA alert)
+    chunkIds.zipWithIndex.foreach {
```
parkan, line 53: does this mean that we have to possibly retry every failed chunk manually :scream:

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
vyzo, line 19: ugh indeed -- we start down that road next thing you know we'll be pulling left-pad :)

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
vyzo: yes, the underlying implementation of the binary datastore.

transactor/src/main/scala/io/mediachain/datastore/BinaryDatastore.scala
vyzo, line 30: I have it returning no useful results so that the Multihash reference construction happens in one place, in the BinaryDatastore implementation.

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
```diff
@@ -0,0 +1,158 @@
...
+      div + 1
+    } else {
+      div
+    }
+  }
+  
+  private def bytes2Buffer(bytes: Array[Byte]) = 
+    ByteBuffer.wrap(bytes)
+  
+  private def bytes2BufferChunk(bytes: Array[Byte], chunk: Int) = {
```
parkan, line 97: `bytes2Buffer(val)` is pretty much `bytes2BufferChunk(val, 0)`, no?

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
vyzo, line 53: no, the intention is to have the PersistentStore that does queued writes retry the entire put if any chunk fails.

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
vyzo, line 97: it is equivalent, but implementing it this way would create an unnecessary slice.

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
```diff
@@ -0,0 +1,158 @@
...
+  override def get(key: MultiHash): Option[Array[Byte]] = {
+    val keyAttr = new AttributeValue
+    keyAttr.setS(key.base58)
+    
+    val item = Map(
+      "multihash" -> keyAttr
+    )
+    val res = db.getItem(table, item).getItem  // yes, really.
+    
+    Option(res).map { item =>
```
parkan, line 116: maybe like this

``` scala
def handleNull[T](x: => T): Option[T] = try Option(x) catch {
  case _: NullPointerException => None
}

handleNull(res.get("data"))
  .getOrElse(handleNull(res.get("chunks")))
  .getOrElse(throw new DatastoreException("Bad record: " + key.base58))
```

or

```
for {
  r <- Option(res)
  data <- Option(r.get("data"))
} yield data
...
```

vyzo, line 116: Actually if data returns null then it looks for chunked representation, so this won't quite work.

transactor/src/main/scala/io/mediachain/datastore/BinaryDatastore.scala
parkan, line 30: it's just weird for a void side-effecty method to have the same name as a method returning a value on the same class, you know?

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
parkan, line 19: you usually get it for free with other twitter utils :)

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
parkan, line 53: cool ok so then exception bailing on the whole thing makes sense

transactor/src/main/scala/io/mediachain/datastore/BinaryDatastore.scala
vyzo, line 30: I see where you are coming from -- I can rename it to something more apt.

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
parkan, line 116: Hmm?

``` scala
scala> val foo: String = null
foo: String = null

scala> handleNull(foo).getOrElse("this would try chunks next")
res9: String = this would try chunks next
```

The function handles a null value as well as NPE

transactor/src/main/scala/io/mediachain/datastore/BinaryDatastore.scala
parkan, line 30: :+1:

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
vyzo, line 116: The logic is that first you check if you have a "data" attribute, which stores the binary directly.
If this is null, then you check for a chunks attribute which is a list of chunk-ids to fetch from the Chunks table.
So the value that you get out of the chunks attribute does not substitute the value of the data attribute for the monadic approach to work.

Also, there should be no case of having a NullPointer exception the way it is written, and I really don't want to swallow NPEs

parkan, line 116: Ah oops I forgot the outer funcs:

``` scala
handleNull(buffer2Bytes(item.get("data").getB))
  .getOrElse(handleNull(getChunks(item.get("chunks").getSS.toList)))
  .getOrElse(throw new DatastoreException("Bad record: " + key.base58))
```

This isn't super sexy either though, fair enough (plus the boxing/unboxing overhead is not zero)

I can live with these literal null checks etc in this code, I guess

vyzo, line 116: This will throw an NPE if it is a chunked representation.
I find the null checks code cleaner and more straightforward to reason about here, since we are interfacing a decisively java api.

parkan, line 116: It would be handled no matter where it happened, but I'm ok keeping your approach

PR #66, opened 2016-05-09
wip - scalacheck generators for blockchain + datastore

yusefnapora:
This adds a `genJournalBlock` function that creates scalacheck generator which returns a `JournalBlock` plus an in-memory "datastore" as a `Map[Reference, DataObject]`.  

You can feed these back into the `genJournalBlock` function, which will generate a new block + data objects, building on the previous one.

Needs some cleanup, and I want to test that the references are valid before using this for the client specs.

### DIFF
--- a/client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
+++ b/client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
@@ -1,0 +1,97 @@
+package io.mediachain.client
+
+import io.mediachain.protocol.Datastore._
+
+class CanonicalFolder(datastore: Datastore) {
+  import io.mediachain.util.cbor.CborAST.CValue
+
+  /**
+    * Tries to extend the list of `cells` by following the `chain`
+    * reference of the cell at the head of the list.  If successful,
+    * prepends the new cell to the head of the list and repeats until
+    * a nil `chain` reference is found, or the referenced cell can't
+    * be retrieved from the datastore.
+    *
+    * @param cells - a list of `ChainCell`s to extend
+    * @tparam T - Should be either EntityChainCell or ArtefactChainCell -
+    *           using a concrete subtype may fail to find all cells.
+    * @return - all the `ChainCell`s that can be found by chasing
+    *         `chain` references from the head of the input list
+    */
+  def extendChain[T <: ChainCell](cells: List[T]): List[T] = {
+    val nextCellRef = cells.headOption.flatMap(_.chain)
+    val nextCellOpt = nextCellRef.flatMap(datastore.getAs[T])
+    if (nextCellOpt.isEmpty) {
+      cells
+    } else {
+      extendChain(nextCellOpt.toList ++ cells)
+    }
+  }
+
+  /**
+    * Given a list of `ChainCell`s, return a folded `meta` map containing
+    * each cell's `meta` map applied in sequence.
+    * @param chain - a list of `ChainCell`s to merge
+    * @return a flattened `meta` map
+    */
+  def foldedMeta(chain: List[ChainCell]): Map[String, CValue] =
+    // FIXME: this is very naive and doesn't flag or otherwise resolve conflicts
+    chain.foldLeft(Map[String, CValue]()) { (meta, cell) =>
+      meta ++ cell.meta
+    }
+
+  /**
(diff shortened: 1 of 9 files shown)


parkan:
I like having the generators in the protocol test namespace -- this gives us reusability (for example, the indexer client will probably need these) but without bloating distribution for protocol too much (should we distribute it)

protocol/src/test/scala/io/mediachain/protocol/CborSerializationSpec.scala
```diff
@@ -217,10 +217,10 @@ object CborSerializationSpec extends BaseSpec with ScalaCheck {
         obj must haveClass[ArtefactChainCell]
       }
     }.setGen(Gen.oneOf(
-        genArtefactReferenceCell,
-        genArtefactUpdateCell,
-        genArtefactOwnershipCell,
-        genArtefactDerivationCell,
-        genArtefactCreationCell
+        genArtefactReferenceCell(),
```
parkan: do these have side effects now?

yusefnapora: I made them into functions with default parameters, instead of just `val`s.  So that you can control what `entity` or `artefact` gets referenced by the cell, etc.

protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
```diff
@@ -33,49 +39,87 @@ object DataObjectGenerators {
...
+    artefactOriginGen: Gen[Reference] = genReference
+  ) = for {
+    base <- genArtefactChainCell(artefactGen, chainGen)
+    artefactOrigin <- artefactOriginGen
   } yield ArtefactDerivationCell(base.artefact, base.chain, base.meta, artefactOrigin)
 
-  val genArtefactOwnershipCell = for {
-    base <- genArtefactChainCell
-    entity <- genReference
+  def genArtefactOwnershipCell(
```
parkan: I'm not sure we need these (i.e. specialized types beyond Update), though that may be a separate discussion. We're getting kind of close to the point where we need to have a serious think about how all this stuff is modeled.

protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
```diff
@@ -0,0 +1,195 @@
...
+    }
+  }
+
+
+  def refForCell(cell: ChainCell): Reference = cell match {
+    case e: EntityChainCell => e.entity
+    case a: ArtefactChainCell => a.artefact
+  }
+
+  def consChainCells(cells: List[ChainCell]): List[ChainCell] = {
```
parkan: can you add a small comment explaining what this does?

yusefnapora: yeah, very good call :)

yusefnapora:
@parkan I made a start on the spec for the client `JournalListener` implementation, but it's still pretty barebones.  I re-animated the `client` sub-project and put it in there, thinking it might be good to have common functionality like that in one place, and l-space and other future clients could depend on it...  what do you think; is that worth the extra project?

I still need to clean up the val vs def thing in the `DataObject` generators we were discussing yesterday, but got too focused on the blockchain generators this morning.  Will do that after lunch

parkan:
Yeah, the `client` core should be reusable for sure (for ex. for the indexer). This will also give us an easy way to migrate everything if/when we move off of copycat protocol.

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
```diff
@@ -0,0 +1,97 @@
...
+      case (a: Artefact, None) => a
+
+      case (e: Entity, Some(c: EntityChainCell)) =>
+        foldedEntity(e, Some(c))
+
+      case (a: Artefact, Some(c: ArtefactChainCell)) =>
+        foldedArtefact(a, Some(c))
+
+      case _ =>
+        throw new IllegalStateException(
```
vyzo, line 91: this is more of an IllegalArgumentException.

yusefnapora:
Going to close this and reopen with just the generators.

Issue #26, opened 2016-04-26
Datastore


parkan:
A black box k/v store that takes the place of IPFS for now. DynamoDB?

vyzo:
#61 is relevant.

bigs:
Initial work is done

PR #68, opened 2016-05-10
Perisistent Datastore implementation

vyzo:
#61 #26

Implements a PersistentDatastore which acts a write-through cache to Dynamo:
data are disk-persisted in a local rocks datastore until they are written through to Dynamo by a background thread.

Further refinements:
- more complete error handling
- perhaps chattier logs to help with debugging

### DIFF
--- a/transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
+++ b/transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
@@ -16,13 +16,12 @@ class PersistentDatastore(config: PersistentDatastore.Config)
   val maxBackoffRetry = 60 // seconds
 
   override def putData(key: MultiHash, value: Array[Byte]) {
-    rocks.synchronized {rocks.putData(key, value)}
+    rocks.putData(key, value)
     queue.putLast(key)
   }
 
   override def getData(key: MultiHash): Option[Array[Byte]] = {
-    rocks.synchronized {rocks.getData(key)}
-      .orElse {dynamo.synchronized {dynamo.getData(key)}}
+    rocks.getData(key).orElse {dynamo.synchronized {dynamo.getData(key)}}
   }
   
   override def close() {
@@ -41,14 +40,14 @@ class PersistentDatastore(config: PersistentDatastore.Config)
   private def recover() {
     // Crash recovery: schedule all keys still in rocks db for
     //  write-through to dynamo
-    rocks.synchronized {rocks.getKeys}.foreach {key => queue.putLast(key)}
+    rocks.getKeys.foreach {key => queue.putLast(key)}
   }
   
   private def loop(backoff: Int) {
     if (!Thread.interrupted) {
       writeNext() match {
         case Some(key) => {
-          rocks.synchronized {rocks.removeData(key)}
+          rocks.removeData(key)
           loop(0)
         }
           
@@ -66,7 +65,7 @@ class PersistentDatastore(config: PersistentDatastore.Config)
   private def writeNext(): Option[MultiHash] = {
     val key = queue.takeFirst
     try {
-      rocks.synchronized {rocks.getData(key)}.map { data =>
+      rocks.getData(key).map { data =>
         dynamo.synchronized {dynamo.putData(key, data)}
         key
       }
--- a/transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
+++ b/transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
@@ -40,8 +40,9 @@ class PersistentDatastore(config: PersistentDatastore.Config)
   }
   
   private def recover() {
-    // XXX Implement me: crash recovery
-    //     enqueue all rocks persisted keys for write through to dynamo
+    // Crash recovery: schedule all keys still in rocks db for
+    //  write-through to dynamo
+    rocks.synchronized {rocks.getKeys}.foreach {key => queue.putLast(key)}
   }
   
   private def loop(backoff: Int) {
--- a/transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
+++ b/transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
@@ -3,11 +3,13 @@ package io.mediachain.datastore
 import java.util.Random
 import java.util.concurrent.LinkedBlockingDeque
 import com.amazonaws.AmazonServiceException
+import org.slf4j.{Logger, LoggerFactory}
 import io.mediachain.multihash.MultiHash
 
 class PersistentDatastore(config: PersistentDatastore.Config)
   extends BinaryDatastore with AutoCloseable with Runnable {
   
+  val logger = LoggerFactory.getLogger(classOf[PersistentDatastore])
   val dynamo = new DynamoDatastore(config.dynamo)
   val rocks = new RocksDatastore(config.rocks)
   val queue = new LinkedBlockingDeque[MultiHash]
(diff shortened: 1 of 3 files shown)


transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -0,0 +1,92 @@
...
+    rocks.putData(key, value)
+    queue.putLast(key)
+  }
+
+  override def getData(key: MultiHash): Option[Array[Byte]] = {
+    rocks.getData(key).orElse {dynamo.getData(key)}
+  }
+  
+  override def close() {
+    writer.interrupt()
```
bigs, line 34: could prob do this w/ a volatile bool flag

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -0,0 +1,92 @@
...
+
+class PersistentDatastore(config: PersistentDatastore.Config)
+  extends BinaryDatastore with AutoCloseable with Runnable {
+  
+  val logger = LoggerFactory.getLogger(classOf[PersistentDatastore])
+  val dynamo = new DynamoDatastore(config.dynamo)
+  val rocks = new RocksDatastore(config.rocks)
+  val random = new Random
+  val maxBackoffRetry = 60 // second
+  val queue = new LinkedBlockingDeque[MultiHash]
```
bigs, line 17: perhaps this should be a `ConcurrentLinkedQueue` as this is an SPSC usecase

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -0,0 +1,92 @@
...
+  }
+  
+  private def recover() {
+    // Crash recovery: schedule all keys still in rocks db for
+    //  write-through to dynamo
+    rocks.getKeys.foreach {key => queue.putLast(key)}
+  }
+  
+  private def loop(backoff: Int) {
+    if (!Thread.interrupted) {
```
bigs, line 58: hmmm didn't realize this was reliable. if so, ignore the bool flag comment

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -0,0 +1,92 @@
...
+class PersistentDatastore(config: PersistentDatastore.Config)
+  extends BinaryDatastore with AutoCloseable with Runnable {
+  
+  val logger = LoggerFactory.getLogger(classOf[PersistentDatastore])
+  val dynamo = new DynamoDatastore(config.dynamo)
+  val rocks = new RocksDatastore(config.rocks)
+  val random = new Random
+  val maxBackoffRetry = 60 // second
+  val queue = new LinkedBlockingDeque[MultiHash]
+  val writer = new Thread(this, "PersistentDataStore#write")
```
bigs: i definitely see what yr getting at with this, though i feel pretty strongly that this should be called from the outside. this is a very surprising side effect to someone from the outside world. if anything, i could see there being an instance method `start()` or something like this to make it more clear so you'd have

``` scala
val pds = new PersistentDatastore(config)
pds.start
```

bigs: also maybe give the thread a unique name like `"PersistentDatastore#${this.hashCode}"`

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -0,0 +1,92 @@
...
+  private def recover() {
+    // Crash recovery: schedule all keys still in rocks db for
+    //  write-through to dynamo
+    rocks.getKeys.foreach {key => queue.putLast(key)}
+  }
+  
+  private def loop(backoff: Int) {
+    if (!Thread.interrupted) {
+      writeNext() match {
+        case Some(key) => {
```
bigs: block is unnecessary, just indentation will suffice

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -0,0 +1,92 @@
...
+          rocks.removeData(key)
+          loop(0)
+        }
+          
+        case None => {
```
bigs: same

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
parkan: Presumably we also want a thread pool

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
vyzo, line 58: yeah, it's reliable -- either an InterruptedException or the flag will be set.

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
vyzo: will add start method and unique name.
@parkan -- why a thread pool? we just need one thread doing the queued writes and it's much easier to manage concurrency with a single worker.

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
vyzo: ok, will remove noise.

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
vyzo: ok, will denoise.

vyzo: which block do you want to remove here? I prefer to keep the one around if by habbit.

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
parkan: @vyzo I'm fairly sure using Futures requires a thread pool (by default the ForkJoinPool from the global ExecutionContext)

Given that the order of writes to the external datastore doesn't really matter as long as they all complete eventually, and that you're retrying/queueing anyway, I don't see why a small number of concurrent writers isn't appropriate.

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
vyzo: okie, killed the blocks in pattern matching.

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
vyzo: Noted. It adds some complexity in the code -- we can add it if the throughput from a single writer thread turns out to be insufficient.

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -0,0 +1,92 @@
...
+    val key = queue.takeFirst
+    try {
+      rocks.getData(key).map { data =>
+        dynamo.putData(key, data)
+        key
+      }
+    } catch {
+      case e: AmazonServiceException =>
+        logger.error("AWS Error writing " + key.base58, e)
+        queue.putFirst(key)
```
parkan, line 84: So if there's a single Bad Item that consistently dies we're gonna short-circuit on it and never make progress, right? Maybe this should be a `putLast`?

PR #69, opened 2016-05-10
Rename root sbt project to make intellij happy

yusefnapora:
IntelliJ's sbt indexer gets confused if the root project is named "root", because it conflicts with the root project of the scala-multihash dependency, which is also named "root".  This doesn't affect compilation by sbt itself, but is annoying when editing, because IntelliJ can't find the multihash project and flags it as missing.

This renames the root project to "mediachain", which causes everything to index correctly.

Also removes the "circle_1" and "circle_2" subprojects, since we're using Travis now.


vyzo:
:+1:

parkan:
oh good pt I forgot the circle projects got left in there

PR #72, opened 2016-05-11
copycat client tidbits

vyzo:
A couple of details in Copycat.Client:
- specify seemingly best connection and recovery strategies available
- provide accessor to the underlying CopycatClient object; this is useful for receiving state change notifications


yusefnapora:
👍 looks good

PR #74, opened 2016-05-11
Scalacheck generators for blockchain + data objects

yusefnapora:
Just the generators from #66 

Adds scalacheck generators for single journal blocks (with generated `DataObject`s in an `InMemoryDatastore`), and a blockchain of linked blocks.

### DIFF
--- a/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
+++ b/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
@@ -1,18 +1,20 @@
 package io.mediachain.protocol
 
-import io.mediachain.multihash.MultiHash
-
 
 object Datastore {
   import io.mediachain.multihash.MultiHash
   import io.mediachain.protocol.CborSerialization._
   import io.mediachain.protocol.Transactor.{ArtefactChainReference, ChainReference, EntityChainReference}
   import io.mediachain.util.cbor.CborAST._
+  import scala.util.Try
 
   // Datastore interface
   trait Datastore {
     def get(ref: Reference): Option[DataObject]
     def put(obj: DataObject): Reference
+
+    def getAs[T <: DataObject](ref: Reference): Option[T] =
+      get(ref).flatMap(obj => Try(obj.asInstanceOf[T]).toOption)
   }
 
   class DatastoreException(what: String) extends RuntimeException(what)
--- a/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
+++ b/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
@@ -1,5 +1,6 @@
 package io.mediachain.protocol
 
+import io.mediachain.multihash.MultiHash
 
 
 object Datastore {
@@ -36,6 +37,12 @@ object Datastore {
       CMap.withStringKeys("@link" -> CBytes(multihash.bytes))
   }
 
+  object MultihashReference {
+    def forDataObject(dataObject: DataObject): MultihashReference =
+      MultihashReference(
+        MultiHash.hashWithSHA256(dataObject.toCborBytes)
+      )
+  }
 
   // Canonical records: Entities and Artefacts
   sealed abstract class CanonicalRecord extends Record {
--- a/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
+++ b/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
@@ -74,6 +74,7 @@ object Datastore {
 
   // Chain cells
   sealed abstract class ChainCell extends Record {
+    def ref: Reference
     def chain: Option[Reference]
     def cons(chain: Option[Reference]): ChainCell
   }
@@ -83,6 +84,8 @@ object Datastore {
     val chain: Option[Reference],
     val meta: Map[String, CValue]
   ) extends ChainCell {
+    val ref = entity
+
     override def cons(xchain: Option[Reference]): ChainCell =
       EntityChainCell(entity, xchain, meta)
 
@@ -111,6 +114,8 @@ object Datastore {
     val chain: Option[Reference],
     val meta: Map[String, CValue]
   ) extends ChainCell {
+    val ref = artefact
+
     override def cons(xchain: Option[Reference]): ChainCell =
       ArtefactChainCell(artefact, xchain, meta)
     
--- a/protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
+++ b/protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
@@ -1,6 +1,7 @@
 package io.mediachain.protocol
 
 import java.nio.charset.StandardCharsets
+import java.util.Date
 
 
 object DataObjectGenerators {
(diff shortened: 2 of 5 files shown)


protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
```diff
@@ -0,0 +1,193 @@
...
+    * @return depending on the type of `CanonicalRecord` given, a generator for
+    *         either a subtype of `EntityChainCell` or `EntityArtefactCell`.
+    *         Note that the `chain` field of the generated cell will be
+    *         set to `None`
+    *
+    */
+  def genCellFor(canonical: CanonicalRecord)
+  : Gen[ChainCell] =
+    canonical match {
+      case e: Entity => genEntityUpdateCell(e, genNilReference)
```
parkan, line 25: inexhuastive match

parkan:
This looks good to me, I'll make some further changes on top of this for data generation but in and of itself it should be good

protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
yusefnapora, line 25: it is?  I thought because `CanonicalRecord` was sealed you only had to match against the two concrete types.

vyzo, line 25: CanonicalRecord is indeed sealed so it's only Entities and Artefacts.

parkan, line 25: my bad, you're right

protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
```diff
@@ -76,6 +86,8 @@ object Datastore {
     val chain: Option[Reference],
     val meta: Map[String, CValue]
   ) extends ChainCell {
+    val ref = entity
```
bigs, line 89: why not name `entity` `ref` instead

protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
```diff
@@ -104,6 +116,8 @@ object Datastore {
     val chain: Option[Reference],
     val meta: Map[String, CValue]
   ) extends ChainCell {
+    val ref = artefact
```
bigs, line 119: same

protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
```diff
@@ -12,71 +13,134 @@ object DataObjectGenerators {
   import io.mediachain.util.cbor.CborAST._
   import io.mediachain.util.cbor.CValueGenerators._
 
+  val genStringMetas: List[Gen[(String, CValue)]] = (1 to 25).toList.map { _ =>
+    for {
```
bigs, line 17: pretty dang low prio, but someth like this would be cool in the future https://gist.github.com/etorreborre/2231981

applicative style will clean up a looooooooooot of these generators

protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
yusefnapora, line 89: that sounds good to me :)

protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
```diff
@@ -12,71 +13,134 @@ object DataObjectGenerators {
...
-  val genEntityChainCell = for {
-    entity <- genReference
-    chain <- genReference
+  def genReferenceFor(canonicalGen: Gen[CanonicalRecord]): Gen[Reference] =
+    for {
+      canonical <- canonicalGen
+    } yield MultihashReference.forDataObject(canonical)
+
+  def genEntityChainCell(
+    entityGen: Gen[Entity],
```
bigs, line 60: feel free to rule out the idea if it's not useful or too out of scope...

what if these accepted `Either[Type, Gen[Type]]` such that these could be used in scenarios where we wanted to fix a value to a default?

bigs, line 60: edited above

yusefnapora, line 60: if you want to fix a value you can use `Gen.const(someValue)`, which always returns the input.

protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
bigs, line 25: curious if the compiler can reason about this hah.

bigs, line 25: noting that this does appear to be exhaustive

protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
parkan, line 60: `Gen.const(thing)` takes care of this ok

protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
parkan, line 17: sure, feel free to make an issue with that specific gist, looks neat but prob not worth reworking right now

protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
```diff
@@ -0,0 +1,193 @@
...
+    * @return the input list, with the `chain` field of each cell updated
+    *         so that it points to the previous cell that refers to the same
+    *         `CanonicalRecord`.
+    */
+  def consChainCells(cells: List[ChainCell]): List[ChainCell] = {
+    import collection.mutable.{Map => MMap, MutableList => MList}
+    val chainHeads: MMap[Reference, ChainCell] = MMap()
+    val consed: MList[ChainCell] = MList()
+
+    cells.foreach { c =>
```
bigs: not super important but this could be a stateless fold w/ accumulator `(Map..., List)` where u just `._2` after folding to pop out your result

or a recursive fn in the `State` monad from cats (prob easiest)

protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
parkan, line 25: yeah it totally can

protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
```diff
@@ -0,0 +1,193 @@
...
+    * @param blockchain the head of a previously generated journal blockchain
+    * @return a generator of the new head of a journal blockchain, and an
+    *         in-memory datastore containing all generated objects (including
+    *         journal blocks)
+    */
+  def genJournalBlock(
+    blockSize: Int,
+    datastore: => InMemoryDatastore = new InMemoryDatastore,
+    blockchain: Option[JournalBlock] = None
+  ): Gen[JournalBlockWithDatastore] = {
```
bigs, line 76: :+1:

bigs:
i only have suggestions about style. looks good :)

protocol/src/test/scala/io/mediachain/protocol/JournalBlockGenerators.scala
parkan: so it would be

``` scala
cells.foldLeft((Map[Reference, ChainCell], List[ChainCell]())) { ((chainHeads, consed), cell) =>
  val canonicalRef = cell.ref
  val head = chainHeads.get(canonicalRef)
    .map(h => MultihashReference.forDataObject(h))
  val consedCell = cell.cons(head)
  (chainHeads += (canonicalRef, consedCell), consed :+ consedCell)
}
```

yusefnapora: that seems better.. I felt kind of gross about the mutable collections, etc.

protocol/src/test/scala/io/mediachain/protocol/DataObjectGenerators.scala
bigs, line 60: oh perfect

PR #78, opened 2016-05-12
Applicative instance for Gen

bigs:


yusefnapora:
this looks awesome 👍

parkan:
I kind of like the relative clarity of the `for` and explicit constructor call but this is probably for the best given the sheer number of methods

PR #80, opened 2016-05-13
PersistentDatastore: multiple write threads

vyzo:
Adds multiple threads for background writing in PersistentDatastore in order to maximize dynamo throughput.


bigs:
LGTM. generally array backed concurrent queues are faster but this seems fine

vyzo:
Well, we need the background threads to block until a new datum becomes available and concurrent queues don't see to offer take in the interface.

vyzo:
Oh, wait BlockingArrayQueue has take/put methods. I can switch the implementation to use it.

vyzo:
hrm, it requires a capacity bound to be set and blocks the caller thread if capacity is exceeded, which is a problem with the semantics expected by the state machine. I'll keep the linked deque.

bigs:
sounds good

parkan:
The recommended way to deal with multithreading is through `ExecutorService` now, rather than calling `Thread` directly (you can do this with or without futures)

PR #82, opened 2016-05-16
Refactor io.mediachain.transactor to io.mediachain.copycat

vyzo:
- moved the package to io.mediachain.copycat
- refactored Client and Server to separate components out of Copycat module (which no longer exists)
- abstracted Transport and added support for SSL/keystore parameters


transactor/src/test/scala/io/mediachain/copycat/DummyContext.scala
```diff
@@ -16,16 +16,16 @@ object DummyContext {
...
-    val client = Copycat.Client.build()
+    val client = Client.build()
     client.connect(address)
     DummyContext(server, client, store, logdir)
   }
   
   def setupLogdir() = {
     import sys.process._
-    val logdir = "mktemp -d".!!
+    val logdir = "mktemp -d".!!.trim
```
parkan, line 28: was this the issue that tg was experiencing?

parkan:
:+1:

I think this makes sense

transactor/src/test/scala/io/mediachain/copycat/DummyContext.scala
vyzo, line 28: not sure what tg was seeing, but this was a facepalm: the logdir had an extra newline in the end and the rm -rf failed quietly because of this. The result: /tmp/tmp.xxx garbage left behind.

PR #83, opened 2016-05-16
wip - seed cluster with generated/saved blockchain

yusefnapora:
Adds a `SeedingCopycatClient` that tries to replay journal entries in sequence, and a `SeeingCopycatClientSpec` that demonstrates that it's broken :)

It will sometimes work if I only try to submit a single operation, although that also fails sometimes for reasons unknown.  However, I suspect my future chaining code might be broken;  it basically folds over the list of JournalEntries and creates a `flatMap` chain of futures, short-circuiting if one returns `Xor.Left`.

### DIFF
--- a/client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
+++ b/client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
@@ -71,9 +71,9 @@ class CanonicalFolder(datastore: Datastore) {
     * @param canonical - the canonical record
     * @param chainHead - the head of the `CanonicalRecord`'s chain of update cells
     * @return - a new `CanonicalRecord` with the metadata from each cell merged in
-    * @throws IllegalStateException if given a `ChainCell` of the incorrect
-    *                               type for the given `CanonicalRecord`, e.g.
-    *                               if given an `Entity` and an `ArtefactUpdateCell`
+    * @throws IllegalArgumentException if given a `ChainCell` of the incorrect
+    *                                  type for the given `CanonicalRecord`, e.g.
+    *                                  if given an `Entity` and an `ArtefactUpdateCell`
     *
     */
   def foldedCanonical(canonical: CanonicalRecord, chainHead: Option[ChainCell]): CanonicalRecord = {
@@ -88,7 +88,7 @@ class CanonicalFolder(datastore: Datastore) {
         foldedArtefact(a, Some(c))
 
       case _ =>
-        throw new IllegalStateException(
+        throw new IllegalArgumentException(
           s"type mismatch between canonical of type ${canonical.getClass.getTypeName}" +
             s" and cell of type ${chainHead.get.getClass.getTypeName}"
         )
--- a/client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
+++ b/client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
@@ -1,0 +1,97 @@
+package io.mediachain.client
+
+import io.mediachain.protocol.Datastore._
+
+class CanonicalFolder(datastore: Datastore) {
+  import io.mediachain.util.cbor.CborAST.CValue
+
+  /**
+    * Tries to extend the list of `cells` by following the `chain`
+    * reference of the cell at the head of the list.  If successful,
+    * prepends the new cell to the head of the list and repeats until
+    * a nil `chain` reference is found, or the referenced cell can't
+    * be retrieved from the datastore.
+    *
+    * @param cells - a list of `ChainCell`s to extend
+    * @tparam T - Should be either EntityChainCell or ArtefactChainCell -
+    *           using a concrete subtype may fail to find all cells.
+    * @return - all the `ChainCell`s that can be found by chasing
+    *         `chain` references from the head of the input list
+    */
+  def extendChain[T <: ChainCell](cells: List[T]): List[T] = {
+    val nextCellRef = cells.headOption.flatMap(_.chain)
+    val nextCellOpt = nextCellRef.flatMap(datastore.getAs[T])
+    if (nextCellOpt.isEmpty) {
+      cells
+    } else {
+      extendChain(nextCellOpt.toList ++ cells)
+    }
+  }
+
+  /**
+    * Given a list of `ChainCell`s, return a folded `meta` map containing
+    * each cell's `meta` map applied in sequence.
+    * @param chain - a list of `ChainCell`s to merge
+    * @return a flattened `meta` map
+    */
+  def foldedMeta(chain: List[ChainCell]): Map[String, CValue] =
+    // FIXME: this is very naive and doesn't flag or otherwise resolve conflicts
+    chain.foldLeft(Map[String, CValue]()) { (meta, cell) =>
+      meta ++ cell.meta
+    }
+
+  /**
+    * Given an `Entity` and the head of its chain, return a "folded"
+    * representation of the `Entity`
+    * @param entity - the canonical `Entity`
+    * @param chainHead - the head of the `Entity`'s chain of update cells
+    * @return - a new `Entity` with the metadata from each cell merged in
+    */
+  def foldedEntity(entity: Entity, chainHead: Option[EntityChainCell]): Entity = {
+    val chain = extendChain(chainHead.toList)
+    Entity(entity.meta ++ foldedMeta(chain))
+  }
+
+
+  /**
+    * Given an `Artefact` and the head of its chain, return a "folded"
+    * representation of the `Artefact`
+    * @param artefact - the canonical `Artefact`
+    * @param chainHead - the head of the `Artefact`'s chain of update cells
+    * @return - a new `Artefact` with the metadata from each cell merged in
+    */
+  def foldedArtefact(artefact: Artefact, chainHead: Option[ArtefactChainCell]): Artefact = {
+    val chain = extendChain(chainHead.toList)
+    Artefact(artefact.meta ++ foldedMeta(chain))
+  }
+
+  /**
+    * Given an `CanonicalRecord` and the head of its chain, return a "folded"
+    * representation of the `CanonicalRecord`
+    * @param canonical - the canonical record
+    * @param chainHead - the head of the `CanonicalRecord`'s chain of update cells
+    * @return - a new `CanonicalRecord` with the metadata from each cell merged in
+    * @throws IllegalStateException if given a `ChainCell` of the incorrect
+    *                               type for the given `CanonicalRecord`, e.g.
+    *                               if given an `Entity` and an `ArtefactUpdateCell`
+    *
+    */
+  def foldedCanonical(canonical: CanonicalRecord, chainHead: Option[ChainCell]): CanonicalRecord = {
+    (canonical, chainHead) match {
+      case (e: Entity, None) => e
+      case (a: Artefact, None) => a
+
+      case (e: Entity, Some(c: EntityChainCell)) =>
+        foldedEntity(e, Some(c))
+
+      case (a: Artefact, Some(c: ArtefactChainCell)) =>
+        foldedArtefact(a, Some(c))
+
+      case _ =>
+        throw new IllegalStateException(
+          s"type mismatch between canonical of type ${canonical.getClass.getTypeName}" +
+            s" and cell of type ${chainHead.get.getClass.getTypeName}"
+        )
+    }
+  }
+}
--- a/client/src/main/scala/io/mediachain/client/MediachainClient.scala
+++ b/client/src/main/scala/io/mediachain/client/MediachainClient.scala
@@ -1,15 +1,10 @@
 package io.mediachain.client
 
-import cats.data.{Xor, XorT}
+import cats.data.XorT
 import io.mediachain.protocol.Datastore._
-import io.mediachain.protocol.Transactor.{JournalError, JournalListener}
-import io.mediachain.transactor.Copycat.{ClientState, ClientStateListener}
-
-import scala.concurrent.{ExecutionContext, Future}
-import io.mediachain.protocol.Datastore._
+import io.mediachain.protocol.Transactor.JournalError
 
 import scala.concurrent.Future
-import scala.util.{Failure, Success}
 
 
 sealed trait MediachainClientEvent
@@ -53,6 +48,7 @@ trait MediachainClient {
   /**
     * Add an event listener to be informed of new and updated
     * mediachain records.
+    *
     * @param listener an event listener
     */
   def addListener(listener: ClientEventListener): Unit
@@ -60,15 +56,17 @@ trait MediachainClient {
 
   /**
     * Add a new CanonicalRecord to the system
+    *
     * @param canonicalRecord a new Entity or Artefact to add to the system
-    * @return an Xor-wrapped future that will complete with either a ClientError or a
-    *         Reference to the new record
+    * @return an Xor-wrapped future that will complete with either a ClientError
+    *         or a Reference to the new record
     */
   def addCanonical(canonicalRecord: CanonicalRecord): XorT[Future, ClientError, Reference]
 
 
   /**
     * Update an existing CanonicalRecord
+    *
     * @param canonicalReference a Reference to the canonical to update
     * @param chainCell a ChainCell containing the new metadata for the record.
     * @return an Xor-wrapped future that will complete with either a ClientError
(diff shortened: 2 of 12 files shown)


client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
```diff
@@ -0,0 +1,155 @@
...
+    * Add a new CanonicalRecord to the system
+    *
+    * @param canonicalRecord a new Entity or Artefact to add to the system
+    * @return an Xor-wrapped future that will complete with either a ClientError or a
+    *         Reference to the new record
+    */
+  def addCanonical(canonicalRecord: CanonicalRecord) =
+    XorT[Future, ClientError, Reference] {
+      cluster.insert(canonicalRecord)
+        .map(_.bimap(ClientError.Journal, _.ref))
```
vyzo, line 38: we are swallowing errors here -- shouldn't a the very least log the actual error?

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+    * Update an existing CanonicalRecord
+    *
+    * @param canonicalReference a Reference to the canonical to update
+    * @param chainCell          a ChainCell containing the new metadata for the record.
+    * @return an Xor-wrapped future that will complete with either a ClientError
+    *         or a Reference to the new head of the record's chain
+    */
+  def updateCanonical(canonicalReference: Reference, chainCell: ChainCell) =
```
vyzo, line 49: same here

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
```diff
@@ -0,0 +1,155 @@
...
+  /**
+    * Request the current block from the transactor cluster, and fetch
+    * any missing blocks from the blockchain.
+    */
+  def catchupJournal(): Unit = {
+    // Note that we don't set the `latestBlockIndex` to the index of the
+    // current block, since the current block is still in the process of
+    // being constructed by the transactor cluster.
+    cluster.currentBlock.map { block =>
+      block.entries.foreach(handleJournalEntry)
```
vyzo, line 102: entries are consumed in the wrong order: previous block should be consumed before the current block's entries.

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
parkan, line 38: the bimap propagates the error

parkan, line 38: oh I see, it should capture the actual error body, yeah

parkan, line 38: so like `_.bimap(err => ClientError.Journal(err), rec => rec.ref)`

yusefnapora, line 38: the ClientError.Journal just wraps a JournalError, so the original is still accessible.

vyzo, line 38: ok, didn't notice that.

yusefnapora, line 38: I like arkadiy's better; it's more explicit about passing the error through.

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
```diff
@@ -0,0 +1,97 @@
...
+    *
+    * @param cells - a list of `ChainCell`s to extend
+    * @tparam T - Should be either EntityChainCell or ArtefactChainCell -
+    *           using a concrete subtype may fail to find all cells.
+    * @return - all the `ChainCell`s that can be found by chasing
+    *         `chain` references from the head of the input list
+    */
+  def extendChain[T <: ChainCell](cells: List[T]): List[T] = {
+    val nextCellRef = cells.headOption.flatMap(_.chain)
+    val nextCellOpt = nextCellRef.flatMap(datastore.getAs[T])
```
bigs, line 23: these two lines should just be a for comprehension

transactor/src/test/scala/io/mediachain/copycat/SeedingCopycatClient.scala
```diff
@@ -0,0 +1,97 @@
...
+
+        canonical.map(c => client.insert(c)
+            .map { res =>
+              println(s"replayed entry ${journalEntry.index}. result: $res")
+              res
+            }
+        )
+          .getOrElse(
+            Future.failed(
+              new RuntimeException("Unable to replay journal entry: " +
```
parkan, line 33: We should be careful about how we mix futures and Xors. I think we can either put the errors in the Xor.left, or in the Future.failed, just as long as we're consistent with it.

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
```diff
@@ -0,0 +1,97 @@
...
+    if (nextCellOpt.isEmpty) {
+      cells
+    } else {
+      extendChain(nextCellOpt.toList ++ cells)
```
bigs, line 27: maybe cons then reverse to avoid n^2'ing

transactor/src/test/scala/io/mediachain/copycat/SeedingCopycatClient.scala
```diff
@@ -0,0 +1,97 @@
...
+
+
+  def seed(blockchain: List[JournalBlock], datastore: Datastore): Future[Xor[JournalError, JournalEntry]] = {
+
+    val journalEntries = blockchain.flatMap(_.entries.toList).sortBy(_.index)
+
+    def empty: Future[Xor[JournalError, JournalEntry]] =
+      Future(Xor.right(CanonicalEntry(index = -1, ref = new DummyReference(-1))))
+
+    def serialize(f1: => Future[Xor[JournalError, JournalEntry]], f2: => Future[Xor[JournalError, JournalEntry]]) =
```
parkan, line 60: I'm really confused about what this is supposed to do

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
```diff
@@ -0,0 +1,97 @@
...
+    *                                  if given an `Entity` and an `ArtefactUpdateCell`
+    *
+    */
+  def foldedCanonical(canonical: CanonicalRecord, chainHead: Option[ChainCell]): CanonicalRecord = {
+    (canonical, chainHead) match {
+      case (e: Entity, None) => e
+      case (a: Artefact, None) => a
+
+      case (e: Entity, Some(c: EntityChainCell)) =>
+        foldedEntity(e, Some(c))
```
vyzo, line 85: why make a new Some object and not just pass chainHead?

transactor/src/test/scala/io/mediachain/copycat/SeedingCopycatClient.scala
yusefnapora, line 60: well... the idea is to flatMap `f2` onto `f1` so that it executes after `f1` completes.  Since `Future.sequence` (which was my first thought) executes all futures in parallel.  In this case it bails out if the result of `f1` is an `Xor.Left`.

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
yusefnapora, line 85: hmm, good point

client/src/main/scala/io/mediachain/client/MediachainClient.scala
```diff
@@ -0,0 +1,80 @@
...
+// so far only the low-level interface is defined, and only References to
+// objects in the datastore are returned.
+// TODO: define a high-level interface that returns CanonicalRecords and
+// ChainCells directly from the datastore
+trait MediachainClient {
+
+  /**
+    * @return the set of References to all known CanonicalRecords
+    */
+  def allCanonicalReferences: Set[Reference]
```
vyzo, line 38: ugh, do we really want to return a billion entries here?

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
parkan, line 27: `++` will be `O(n)` here but `n=1` (or 0) so we're fine

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
```diff
@@ -0,0 +1,155 @@
...
+
+class MediachainCopycatClient(datastore: Datastore)
+  (implicit executionContext: ExecutionContext = ExecutionContext.global)
+  extends MediachainClient with JournalListener with ClientStateListener
+{
+
+  def allCanonicalReferences = canonicalRefs
+
+  def chainForCanonical(ref: Reference): Future[Option[Reference]] = {
+    // TODO: handle disconnected cluster state
```
vyzo, line 19: if the client state is disconnected this will fail the Future immediately.

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
parkan, line 38: yeah, I should have read more closely: it would have to be `_ => ClientError.Journal` in order to actively discard the error, but that doesn't jump out at you

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
parkan, line 102: so we want to recursively build up a lazy list and consume it backwards, maybe?

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+    * Given an `Entity` and the head of its chain, return a "folded"
+    * representation of the `Entity`
+    * @param entity - the canonical `Entity`
+    * @param chainHead - the head of the `Entity`'s chain of update cells
+    * @return - a new `Entity` with the metadata from each cell merged in
+    */
+  def foldedEntity(entity: Entity, chainHead: Option[EntityChainCell]): Entity = {
+    val chain = extendChain(chainHead.toList)
```
bigs, line 51: niiiiiiiiiiiiiiiiiiiiiiiiiiice

client/src/main/scala/io/mediachain/client/MediachainClient.scala
parkan, line 38: this should probably be a `Stream`

yusefnapora, line 38: yeah, for sure.

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
parkan, line 19: We can either block or (my preferred way) reschedule this future by flatMapping onto a future that represents reestablishing the connection. You can put timeouts on that etc too if you want a "hard" failure eventually

transactor/src/test/scala/io/mediachain/copycat/SeedingCopycatClient.scala
```diff
@@ -0,0 +1,97 @@
...
+    val p = Promise[Xor[JournalError, JournalEntry]]()
+    client.addStateListener(new ClientStateListener {
+      override def onStateChange(state: ClientState): Unit = {
+        println(s"*** client state changed: $state")
+//        val serialOps = journalEntries.foldLeft(empty) {
+//          (accumF, entry) => serialize(accumF, replayEntry(entry, client, datastore))
+//        }
+//        p.completeWith(serialOps)
+
+        val headOp = replayEntry(journalEntries.head, client, datastore)
```
vyzo, line 75: Be aware that you are holding up a copycat client thread here.  You want to run this in an executor task, as replayEntry blocks on datastore access.

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
parkan, line 23: This does read easier

``` scala
for {
  thisCell <- cells.headOption
  nextCellRef <- thisCell.chain
  nextCellOpt <- dataStore.getAs[T](nextCellRef) // note: this should probably be a Future[Option...
  nextCell <- nextCellOpt
} yield nextCell
```

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
```diff
@@ -0,0 +1,97 @@
...
+    * a nil `chain` reference is found, or the referenced cell can't
+    * be retrieved from the datastore.
+    *
+    * @param cells - a list of `ChainCell`s to extend
+    * @tparam T - Should be either EntityChainCell or ArtefactChainCell -
+    *           using a concrete subtype may fail to find all cells.
+    * @return - all the `ChainCell`s that can be found by chasing
+    *         `chain` references from the head of the input list
+    */
+  def extendChain[T <: ChainCell](cells: List[T]): List[T] = {
```
parkan, line 21: this needs `@tailrec`?

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
vyzo, line 19: yeah, the only reasonable reaction to disconnecting is to complain loudly and back-off retry. And if you keep failing for too long give up altogether.

vyzo, line 19: careful with the thread on which you call Client#connect as its implementation is blocking.

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
bigs, line 38: could somewhat elegantly solve this with a static method `ClientError.Journal.logAndCreate`

transactor/src/test/scala/io/mediachain/copycat/SeedingCopycatClient.scala
parkan, line 75: we should have two separate pools, one for dealing with the network/copycat and one for dealing with applying changes locally

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+    * Request the current block from the transactor cluster, and fetch
+    * any missing blocks from the blockchain.
+    */
+  def catchupJournal(): Unit = {
+    // Note that we don't set the `latestBlockIndex` to the index of the
+    // current block, since the current block is still in the process of
+    // being constructed by the transactor cluster.
+    cluster.currentBlock.map { block =>
```
bigs, line 101: `foreach`

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
parkan, line 21: I would also maybe break up `cells` into a `cells` accumulator and `nextCellOpt` param, rather than passing it together

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
parkan, line 51: yeah I'm feeling this logic

client/src/test/scala/io/mediachain/client/InMemoryJournalFollower.scala
```diff
@@ -0,0 +1,108 @@
...
+import scala.collection.immutable.SortedMap
+
+class InMemoryJournalFollower(val datastore: Datastore) extends JournalListener {
+
+  override def onJournalCommit(entry: JournalEntry): Unit = {
+    // TODO
+  }
+
+  override def onJournalBlock(ref: Reference): Unit =
+    datastore.getAs[JournalBlock](ref).foreach { block =>
```
bigs, line 15: this will never insert a new entry

client/src/test/scala/io/mediachain/client/InMemoryJournalFollower.scala
```diff
@@ -0,0 +1,108 @@
...
+
+
+  private var blockRefMap: SortedMap[BigInt, Reference] = SortedMap()
+
+  def blocks: List[JournalBlock] =
+    blockRefMap.values.toList.flatMap(datastore.getAs[JournalBlock])
+
+  def currentBlock: Option[JournalBlock] =
+    blockRefMap
+      .lastOption
```
bigs, line 27: is this ordered?

client/src/test/scala/io/mediachain/client/InMemoryJournalFollower.scala
vyzo, line 15: another point here is that you are blocking again in an event dispatch thread

client/src/test/scala/io/mediachain/client/InMemoryJournalFollower.scala
```diff
@@ -0,0 +1,108 @@
...
+      .flatMap(pair => datastore.getAs[JournalBlock](pair._2))
+
+  // Map of canonical reference to most recent ChainEntry
+  // for that canonical
+  private var chainHeads: Map[Reference, ChainEntry] = Map()
+
+  /// Set of all `CanonicalRecord`s referenced in the blockchain
+  def canonicals: Set[CanonicalRecord] =
+    blocks.toSet.flatMap { block: JournalBlock =>
+      val refs = block.entries.collect {
```
bigs, line 37: TIL `collect`

client/src/test/scala/io/mediachain/client/InMemoryJournalFollower.scala
```diff
@@ -0,0 +1,108 @@
...
+       case e: CanonicalEntry => e.ref
+      }
+      refs.flatMap(datastore.getAs[CanonicalRecord])
+    }
+
+
+  /// Set of all chain cells in the block chain.
+  def chainCells: Set[ChainCell] =
+    blocks.toSet.flatMap { block: JournalBlock =>
```
bigs, line 46: `for` comprehension

client/src/test/scala/io/mediachain/client/InMemoryJournalFollower.scala
parkan, line 37: edit: never mind, I forgot about another layer of opt

client/src/main/scala/io/mediachain/client/MediachainCopycatClient.scala
vyzo, line 38: Another point here -- do we really want to wrap this in a different kind of error?
Why not use the JournalError directly? It is defined in the Transactor types after all.

I see yusef also wants to have a NetworkError, but perhaps we should extend the error hierarcy in Transactor instead of wrapping.

yusefnapora, line 38: I think extending the error hierarchy is better, yeah.  Perhaps we should define the error types in the protocol project, with a base `MediachainError` type?

transactor/src/test/scala/io/mediachain/copycat/SeedingCopycatClient.scala
parkan, line 60: Ok, so this is related to my comment above -- you want the chain to fail immediately if an Xor.Left is seen at any point, with failure semantics conveyed by Xor rather than Future. You can just use XorT here

client/src/main/scala/io/mediachain/client/CanonicalFolder.scala
bigs, line 85: i was thinking this too, but then it occurred to me that there might be some type enrichment happening here? could maybe do 

``` scala
case (e: Entity, c@Some(_: EntityChainCell)) => foldedEntity(e, c)
```

client/src/main/scala/io/mediachain/client/MediachainClient.scala
bigs, line 38: `cats.Streaming` is p cool

parkan:
are we abandoning this?

PR #84, opened 2016-05-16
Sharing ideas sketch.

autoencoder:


rfc/drafts/indexer-ideas-v1.md
```diff
@@ -0,0 +1,303 @@
...
+  - System allows user to focus on just one of those detected objects in their search / dedupe query.
+  - Relevant to the search / dedupe architecture, because this likely would require for those multiple objects
+    to be indexed separately, along with their location in the media.
+
++ Non-monetary incentives for resource providers? I.e. User donates resources because he likes particular content.
+  - Force resource providers to support all content equally (Freenet).
+  - Allow full control over the content you're supporting (Bittorrent).
+  - Something in the middle (Tor exit nodes)?
+
++ Resource usage considerations?
```
parkan, line 36: This is an important but out of scope discussion: let's assume fully cooperating peers where we have choice of shard allocations for now. Broadly speaking, what we want is a pattern of contributions that mirrors real-world relationships, e.g. indexers of art from public museums might cooperate etc.

rfc/drafts/indexer-ideas-v1.md
```diff
@@ -0,0 +1,303 @@
...
+      - mashups containing this picture
+  - Or, we start with a small number of specific types of semantic similarity to start.
+  - Range of ways to support this, some relatively simple.
+
++ Eviction of images from the index?
+
++ Co-location of images that should be stored together, for faster retrieval / bulk deletion / cache promotion or eviction?
+
+
+## V1 SEARCH / DEDUPE API:
```
parkan, line 63: This sounds like the indexer instance will be bridged to the p2p network through REST (i.e. will not embed a syndication client). I'm guessing this is to make training easier and keep the logic isolated, is that your thinking?

rfc/drafts/indexer-ideas-v1.md
```diff
@@ -0,0 +1,303 @@
...
+
+   - See the recent e.g. CVPR paper on GPU-based ANN lookup.
+   - 1 GPU = ~300x CPU cores.
+   - When immediate results are required, limited by dataset size due to GPU RAM.
+   - Only worry about failure of a single machine, instead of many. Reduced 95th-percentile worse-case response times?
+
+Typical distributed approach is something like the "shards & replicas" approach of ElasticSearch / Lucene:
+
+   - Each shard sees a subset of the data.
+   - Each replica set contains enough shards to fit entire dataset.
```
parkan, line 128: so with data sharded N ways, each set contains M >= N shards with at least one each of 0..N

can we implement this directly on top of ES so we can reuse their orchestration/failover logic? (not that it's super great, but I want to be solving as few hard problems at once as possible)

rfc/drafts/indexer-ideas-v1.md
```diff
@@ -0,0 +1,303 @@
...
+  - Accepting explicit user feedback signals(?)
+  - Mitigating attacks: resource exhaustion, bad feedback, high-trust actors gone bad, etc.
+  - Other possible ideas, see below.
+
+
+## SCALING ANN INDEXES:
+
+Distributed computing for approximate nearest neighbor (ANN) indexes requires some special consideration.
+
+Sharding:
```
parkan, line 110: How does shard key work? Are we partitioning records, hyperspheres, etc?

rfc/drafts/indexer-ideas-v1.md
autoencoder, line 128: Yes, was thinking to use ES and its supporting tools for ANN lookup / text indexing in V1.

Seems there would be a smooth iteration path, from a simple ES-backed solution, where ES does most of the work -> swapping in custom vectorization and re-ranking models, with ES just being used for ANN indexing and lookup -> finally swapping out ES for something more custom suited for the P2P role.

Could maybe consider another step in this iteration, where each distributed node would hold the entire search index. I'll run a quick benchmark to get a feel for the RAM it takes for 1 indexing node to handle e.g. 1MM, 50MM, 1B images. MMAPing large indexes is also possible, but I'd approach that cautiously for time-sensitive queries.

parkan, line 128: Cool, I think this would be good to describe in more detail in the next version of this doc

rfc/drafts/indexer-ideas-v1.md
autoencoder, line 63: Any ideas on how exactly the real-time search system will be connected to the network? Presumably, P2P tracks the addresses of search indexes, and then a direct connection is established between the searcher and the search system?

The dedupe mode could presumably be done in batches and communicate duplicate decisions via the chain. Or, would it make any sense to consider a lazy just-in-time dedupe? (Probably not.)

Isolated components seems to make sense. In the beginning - because several OTC components are being tied together. In the medium term - as we may want to be able to easily swap out components several times as we iterate. It's a highly active research field. In the long run - when moving the components to separate dedicated hardware could be more efficient.

Could be good reasons to embed everything more tightly, though. If so, interested to hear those ideas.

parkan, line 63: I had something like this in mind, but I'm definitely not committed to it

```
+---------------+
|  Transactors  <-----+
+------+--------+     |
       |              |
   copycat/gRPC     dedup
       |              |
+------v--------+     |
|   Indexers    +-----+
+---------------+
|               |
|  Search API   |
|               |
+----^---+------+
     |   |
     REST?
     |   |
 +---+---v-----+
 |Search client|
 +-------------+
```

So indexers receive updates directly from a copycat "client" component and potentially submit merges for duplicates back up, while on the other side they offer a REST API for web services or whatever. Definitely not married to this design though, and 100% with you on keeping things as modular as possible.

I think what may be helpful is to consider the lifecycle in more detail

rfc/drafts/indexer-ideas-v1.md
```diff
@@ -0,0 +1,303 @@
...
+
+
+## GENERAL MC QUESTIONS / IDEAS:
+
++ In addition to edges that support a relationship, also support edges that deny a relationship?
+  - E.g. `NotArtefactDerivationCell`, `NotArtefactCreationCell`, etc.
+  - Or, adding an "invert_relationship" field for the EntityLinkCell. Not to be confused with
+    inverting the direction of the relationship.
+  
++ Support a denser format for specifying edges, other than just one pair at a time?
```
parkan, line 18: This + above could use an example

parkan:
Good ideas, Tim! I think you can restructure pretty much what's already in here a bit more like this for clarity
- more detailed description of V1, including examples and lifecycle description (train, ingest, search in different context, dedup, etc)
- summary of V2 enhancements and discussion of difference in capabilities (this may actually be more than one feature bucket)
- misc features under question like eviction, Xor operators, etc

rfc/drafts/indexer-ideas-v1.md
autoencoder, line 110: In the initial distributed approach, sharding would pseudo-randomly distribute the data evenly, among all shards in a replica set, e.g. by a simple consistent hashing function, or by something even simpler. In this version, all shards in a replica set would see all queries.

In the next iteration, the aim would be to have the shards each specialize in a local area of the search space, and queries would be routed to shards accordingly. So e.g. "random projection trees" would be a good starting basis for this routing. More advanced approaches typically combine several methods, possibly choosing methods based on the data, e.g. construct a small number of random projection trees, then do neighbor of neighbors search for each node in the tree.

Some complications arise from drift typical types of media content as the dataset grows, and rebalancing for that. 

I'll add more algorithm specifics for the RFC doc, for all of these sections.

rfc/drafts/indexer-ideas-v1.md
autoencoder, line 63: Right, was thinking similar for the mode where new media is ingested into the search index. I'll clarify the diagrams a bit.

vyzo, line 63: What's the plan for deduping artefacts that are already in the chain? 
Do we want an ArtefactLinkCell to connect them?

autoencoder, line 63: Yes, seems to me that `ArtefactLinkCell` is a missing piece that we may need.

Perhaps we should create either a table / list of pseudo-code function definitions, for all the graph edge / node operations, along with their currently-known required and optional arguments?

A few questions I'm have after looking at the code / RFC1 - 

```
  - entity $roundTripEntity
  - artefact $roundTripArtefact
  - entity chain cell $roundTripEntityChainCell
  - entity update cell $roundTripEntityUpdateCell
  - entity link cell $roundTripEntityLinkCell
  - artefact chain cell $roundTripArtefactChainCell
  - artefact update cell $roundTripArtefactUpdateCell
  - artefact creation cell $roundTripArtefactCreationCell
  - artefact derivation cell $roundTripArtefactDerivationCell
  - artefact ownership cell $roundTripArtefactOwnershipCell
  - artefact reference cell $roundTripArtefactReferenceCell
  - canonical journal entry $roundTripCanonicalEntry
  - chain journal entry $roundTripChainEntry
  - journal block $roundTripJournalBlock
```

Interpreting those:
- ArtefactCreation - create entity (what about artefact?) node.
- EntityUpdate / ArtefactUpdate - update entity or artefact node.
- EntityLink - create edge between (entity <-> entity), as specified by "relationship".
- ArtefactOwnership - create undirected edge between (entity <-> artefact).
- ArtefactDerivation - create directed edge from (artefact --> artefact).
- ArtefactReference - create undirected edge from (artefact <-> external_identifier).

So, questions:
- There's ArtefactCreation but no EntityCreation?
- EntityLink allows you to create graph edges between 2 entities, with an arbitrarily specified "relationship" string, but other edge types (ArtefactOwnership, ArtefactDerivation, ArtefactReference) are defined in a different way?
- Undirected (artefact <-> artefact) links? E.g. dedupe artefacts, e.g. ArtefactLinkCell?
- Undirected (entity <-> entity) links? E.g. dedupe entites?
  - Are 2 different types of of artist dedupe needed? One where the exact same persona is present twice, and one where an artist is using multiple personas? 
-  Are entities intended only for artists, or to also represent users in the system (who will never have artefact ownership associated with them)?
- Can edges (ArtefactOwnership, ArtefactDerivation, ArtefactReference, EntityLink) receive updates, or only nodes (entities, artefacts)? Does the user just create an object again at a later point in the journal in order to update it?

vyzo, line 63: Don't get confused by the test code.
Entities don't need a creation cell; they can represent physical persons, organizations, users, etc.

parkan, line 63: We should have a separate discussion about the cell taxonomy -- I think having some real data on hand can motivate this. Let's focus on the `Artefact[Creation|Update]Cell` and a generic "link" cell for the moment (though you make a good point about uni- vs bidirectional links)

In my mind, entities are a general purpose class for representing actors in the system, be they artists/creators, organizations signing statements, users, etc but I'm open to arguments against this. We will likely generate blockstack/onename style personas for these.

Links can receive updates, but probably not as a first class entity -- rather as an overriding update on the parent object.

parkan:
superseded by #96, I presume

Issue #85, opened 2016-05-17
Verify transactor performance


parkan:
We're looking to import ~1MM records for a [pilot project](https://github.com/mediachain/mediachain/milestones/Getty%20import)

We should establish performance and stability characteristics of our test cluster in preparation. The generators from #83 can probably help generate the test data.

vyzo:
I think we should make a performance suite that collects baseline measurements, and run it with multiple client-server configurations. 

I can do small scale testing on my laptop -- we can use a large ec2 instance for more.

vyzo:
#89 implements basic testing client-server apps.
Running the single-stepping chain building test on my (not terribly fast) laptop yields 40s for 1000 writes and 195s for 5000 writes.

So ingesting 1M records at this pace would require 11hrs.

vyzo:
Running two clients in parallel each doing 1000 writes gave 35-40s total time for both of them.
So there is a lot of room for performanace improvement.

vyzo:
Added a concurrent write test, which shows the effect of parallelism in performance.
It takes 50s for 5000 writes with 5 concurrent chains and 65s for 10k writes with 10 concurrent chains.

vyzo:
What these numbers indicate is that it should be possible to ingest 1M records in less than 2hrs.

vyzo:
20k writes with 20 concurrent chains takes 1m26.142s, there is substantial room for performance with parallelism.

vyzo:
100k writes with 100 concurrent chains takes 7m14.478s
I think this hits the parallelism limit on my laptop.

vyzo:
the datastore write backlog takes an additional 10-15min to flush to the local dynamo db.

parkan:
Perfect! These numbers are in the ballpark that we need

PR #89, opened 2016-05-17
transactor: assembly and simple client/server components for perf testing

vyzo:
#85
- appropriate band-aids for sbt-assembly in transactor
- io.mediachain.transactor.JournalServer which runs a copycat server node
- io.mediachain.transactor.TestJournalClient for simple perf-testing copycat write performance

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
@@ -1,0 +1,59 @@
+package io.mediachain.transactor
+
+import java.util.Properties
+import java.io.FileInputStream
+import com.amazonaws.auth.BasicAWSCredentials
+import io.atomix.catalyst.transport.Address
+import io.mediachain.copycat.Server
+import io.mediachain.datastore.{PersistentDatastore, DynamoDatastore}
+import scala.collection.JavaConversions._
+import sys.process._
+
+object JournalServer {
+  def main(args: Array[String]) {
+    if (args.length < 1) {
+      println("Arguments: server-config-file [cluster-address ...]")
+      System.exit(1)
+    }
+    
+    val propfile = args.head
+    val cluster = args.tail.toList
+    val props = new Properties
+    props.load(new FileInputStream(propfile))
+    run(props, cluster)
+  }
+  
+  def run(conf: Properties, cluster: List[String]) {
+    def getq(key: String): String =
+      getopt(key).getOrElse {throw new RuntimeException("Missing configuration property: " + key)}
+    
+    def getopt(key: String): Option[String] =
+      Option(conf.getProperty(key))
+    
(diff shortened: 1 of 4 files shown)


transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Properties
+import java.io.FileInputStream
+import com.amazonaws.auth.BasicAWSCredentials
+import io.atomix.catalyst.transport.Address
+import io.mediachain.copycat.Server
+import io.mediachain.datastore.{PersistentDatastore, DynamoDatastore}
+import scala.collection.JavaConversions._
+import sys.process._
+
+object JournalServer {
```
bigs, line 12: `extends App` is the idiomatic way of declaring a main

bigs:
lgtm

transactor/src/main/scala/io/mediachain/transactor/TestJournalClient.scala
```diff
@@ -0,0 +1,70 @@
...
+      } else {
+        println(n)
+        Future {}
+      }
+    }
+    loop(0)
+  }
+  
+  def fail(what: String): Future[Unit] =
+    Future {throw new RuntimeException(what)}
```
parkan: you should use `Future.failed { new RuntimeException(what) }`, which directly gives you a nicely package `Future` object instead of creating an execution context, throwing, catching, etc.

transactor/src/main/scala/io/mediachain/transactor/TestJournalClient.scala
```diff
@@ -0,0 +1,70 @@
...
+    Await.result(op, Duration.Inf)
+  }
+  
+  def buildChain(client: Client, ref: Reference, chainlen: Int) = {
+    def loop(n: Int): Future[Unit] = {
+      if (n < chainlen) {
+        if ((n % 1000) == 0) {
+          println(s"${n}...")
+        }
+        client.update(ref, ArtefactChainCell(ref, None, Map()))
```
parkan, line 57: Let's use `DataObjectGenerators.genArtefactChainCell`?

transactor/src/main/scala/io/mediachain/transactor/TestJournalClient.scala
vyzo: ah nice, there is already a method that does what I want; will use.

transactor/src/main/scala/io/mediachain/transactor/TestJournalClient.scala
vyzo, line 57: I would prefer to have a separate test that handles that -- i'd rather use this test as baseline.

parkan, line 57: wfm

PR #91, opened 2016-05-18
transactor: some more logging

vyzo:
Small one:
Adds debug logging for PersistentDatastore operations and info logging for block generation in JournalStateMachine.
Additional debug logging seems unnecessary in the JournalStateMachine as copycat has extensive logging around state machine operations at debug level.


bigs:
lgtm :hurtrealbad: :+1:

PR #93, opened 2016-05-19
Copycat recovery enhancements

vyzo:
- Adds a background executor for recovery tasks to copycat Client
- Joins the recovery completable future when recovery is initiated for Suspended sessions
- Throws away copycat clients that appear to be fubared and makes fresh ones (reverted)

### DIFF
--- a/transactor/src/main/scala/io/mediachain/copycat/Client.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/Client.scala
@@ -131,7 +131,7 @@ class Client(client: CopycatClient) extends JournalClient {
         } catch {
           case e: InterruptedException => ()
           case e: Throwable =>
-            logger.error("Session Recovery failed", e)
+            logger.error("Session recovery failed", e)
         }
       }})
   }
--- a/transactor/src/main/scala/io/mediachain/copycat/Client.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/Client.scala
@@ -31,7 +31,6 @@ class Client(client: CopycatClient) extends JournalClient {
   private var listeners: Set[JournalListener] = Set()
   private var stateListeners: Set[ClientStateListener] = Set()
   private val logger = LoggerFactory.getLogger(classOf[Client])
-  private val random = new Random
   private val timer = new Timer(true) // runAsDaemon
   private val maxRetries = 5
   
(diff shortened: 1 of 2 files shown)


vyzo:
Successful session recovery here:
https://api.travis-ci.org/jobs/131626453/log.txt?deansi=true

Note that the errors in the cluster spec end are caused because of timeouts leading the cluster to shutdown

transactor/src/main/scala/io/mediachain/copycat/Client.scala
```diff
@@ -121,21 +122,37 @@ class Client(client: CopycatClient) extends JournalClient {
...
   
+  private def recover() {
+    val cf = client.recover()
+    recoveryExecutor.submit(new Runnable {
+      def run {
+        try {
+          cf.join()
+          logger.info("Copycat session recovered")
+        } catch {
+          case e: InterruptedException => ()
```
parkan: The docs say `join` doesn't throw `InterruptedException` (unlike `get`)

parkan: If you're trying to catch a an `InterruptedException` inside the future that you've joined to, you need to catch `CompletionException` and look at underlying with `getCause`

vyzo: ok, I will skip the InterruptedException catch - I am fine with logging it if the client happens gets closed while recovering.

vyzo:
merging as the recovery logic seems to work properly and there are otherwise code improvements.

transactor/src/main/scala/io/mediachain/copycat/Client.scala
parkan: Just to be clear: the call to `cf.join()` can only throw two types of exceptions, `CompletionException` and `CancellationException`. If something inside the future that `cf` encompasses throws any `Exception`, then `join` will throw a  `CompletionException` with that exception accessible through `getCause`. 

This is by contrast to `.get`, which throws `ExecutionException` instead of `CompletionException` but with almost the same semantics, and `InterruptedException` as a near-equivalent of `CancellationException`. I think. This API is super confusing.

vyzo: yeah, i got it. 
It doesn't matter all that much with the interrupt exception, the worst thing that can happen is some errant logging at shutdown.

PR #94, opened 2016-05-19
rfc-1: schema revision

vyzo:
Per our discussion and implementation experience so far:
- store data object metadata in a meta field
- normalize reference field in chain cells
- add ArtefactLinkCells for dedup/similarity linking of artefacts

### DIFF
--- a/rfc/mediachain-rfc-1.md
+++ b/rfc/mediachain-rfc-1.md
@@ -62,12 +62,13 @@ by Mediachain. They can be images, video, text, or any other
 common media found in the Internet. 
 
 In Mediachain schema, entities are instances of data objects with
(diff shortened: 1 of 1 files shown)


parkan:
This looks good and actually upon another reading, maybe some of these specialized cell types actually do make sense.

parkan:
Forgot one thing: we also need a "derived_from" field to point to a raw metadata blob that the update cell was sourced from

vyzo:
there is already an artefactOrigin field that points to the original canonical record and there is the meta that can contain additional metadata, do we really need another field?

parkan:
ah, wasn't sure what the semantics of `artefactOrigin` were, that's fine

PR #97, opened 2016-05-23
reproduce the deadlock

yusefnapora:
@vyzo this adds a `Deadlocker` client program to the transactor project.  It's very braindead - it just constantly creates new artefacts in a background thread, and awaits an infinite flatMap chain of requests for the current block.

I've been testing it against a local server, but I think my connection to the local dynamo is messed up; I'm getting a bunch of these: 

```
INFO: Unable to execute HTTP request: Unrecognized SSL message, plaintext connection?
javax.net.ssl.SSLException: Unrecognized SSL message, plaintext connection?
    at sun.security.ssl.InputRecord.handleUnknownRecord(InputRecord.java:710)
    at sun.security.ssl.InputRecord.read(InputRecord.java:527)
    at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:973)

...
```

But anyway, it will reliably lock up for me, and interestingly, it always does so after either block #507 or 508 - no idea why, or if it will behave the same on your machine.  When it does fail, the background thread keeps successfully creating blocks, but the request for the current block doesn't retry; it just hangs.

I've been testing with the assembled jar using `java -cp ./transactor/target/scala-2.11/transactor-assembly-0.0.1.jar io.mediachain.transactor.Deadlocker 127.0.0.1:54321`

adding `-Dorg.slf4j.simpleLogger.defaultLogLevel=debug` results in mountains of logs.

I'll try to make this nicer in the morning, but please let me know if you find anything 😄

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/CurrentBlock.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/CurrentBlock.scala
@@ -1,0 +1,23 @@
+package io.mediachain.transactor
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+import io.mediachain.copycat.Client
+import io.mediachain.protocol.Datastore.JournalBlock
+
+object CurrentBlock {
+  def main(args: Array[String]) {
+    if (args.length != 1) {
+      println("Expected arguments: server-address")
+      System.exit(1)
+    }
+
+    val server = args(0)
+    val client = Client.build()
+    client.connect(server)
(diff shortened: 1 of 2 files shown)


parkan:
@yusefnapora how can I run the server to try this out against? We should put the invocation in the readme

vyzo:
@yusefnapora you need to specify "http://localhost:8000" (with http prefix) as the endpoint to avoid the dynamo error message.
@parkan you run JournalServer with a properties file that specify its config.

yusefnapora:
closing in favor of #98

PR #98, opened 2016-05-24
Copycat debugging / test programs

yusefnapora:
Apps to test and reproduce client deadlock issues.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/copycat/StateMachine.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/StateMachine.scala
@@ -40,7 +40,7 @@ object StateMachine {
   class JournalState extends Serializable {
     var seqno: BigInt = 0                                     // next entry index
     var index: MMap[Reference, ChainReference] = new MHashMap // canonical -> chain mapp
-    val block: Buffer[JournalEntry] = new ArrayBuffer         // current block entries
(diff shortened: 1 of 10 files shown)


parkan:
Things to try:
- [ ] remove writer activity, prefab a v large block to return, attempt to repro failure case w/o writing
- [ ] synchronize state vars
- [ ] heapdump at beginning and at failure, look for leaks
- [ ] look for any blocks that swallow exceptions

transactor/src/main/scala/io/mediachain/copycat/StateMachine.scala
```diff
@@ -130,10 +130,28 @@ object StateMachine {
...
+        val ref = MultihashReference(multihash = MultiHash.hashWithSHA256(s"foo#$i".getBytes))
+        CanonicalEntry(i, ref)
+      }
+      entries.toArray
+    }
+
     def currentBlock(commit: Commit[JournalCurrentBlock]) : JournalBlock = {
       try {
-        JournalBlock(state.seqno, state.blockchain, state.block.toArray)
+        JournalBlock(state.seqno, state.blockchain, dummyBlock)
```
vyzo: uhm, do we want to keep this in the branch?
We'll have to revert this eventually anyway.

vyzo:
the bug is slain and copycat/catalyst are updated; merging.

PR #99, opened 2016-05-26
Data Model and CBOR serialization updates for rfc-1 schema revisions

vyzo:
- Adds ArtefactLinkCell
- Aggregate record meta data go to a "meta" field in CBOR
- ChainCells have a common "ref" for canonical references

### DIFF
--- a/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
+++ b/protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
@@ -33,7 +33,7 @@ object Datastore {
 
   // Content-addressable reference using IPFS MultiHash
   case class MultihashReference(multihash: MultiHash) extends Reference {
-    val mediachainType: Option[MediachainType] = None
+    override val mediachainType: Option[MediachainType] = None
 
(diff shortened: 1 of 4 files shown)


protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
```diff
@@ -95,12 +101,6 @@ object Datastore {
       meta.get("type")
```
yusefnapora: Will this still work?  Now that `meta` is an explicit field, it shouldn't contain the type name... I added this so that we could deserialize a subtype to the base type but still preserve the original type name in case we re-serialize it later...

vyzo: uhm no, I missed that; we no longer have a raw CMap of the blob -- good catch.

vyzo: ok, got rid of it altogether -- all subtypes correctly define their type appropriately and the user has no business specifying the type anyway.

PR #100, opened 2016-05-27
CBOR serialization for JournalBlocks in Copycat

vyzo:
Blocks serialized with CBOR are half-size compared to their POJO form, so this adds a custom Serializer for them in copycat.
In order to do this and still have tests with dummy references, the DummyReference becomes a first-class citizen of protocol-land.


yusefnapora:
👍 this looks good to me

Issue #29, opened 2016-04-27
Glossary [TODO]


parkan:
We have some inconsistent term usage again, let's nip that in the bud and establish definitions. I am particularly concerned about:

**canonical:** we've generally used this for the root ID but now we seem to also be using it for the multihash IPFS address of an object
**peer/client:** in the RFC we have Peers (~= transactors?) but usage seems to have shifted to peers meaning transactors + nodes accepting syndication + ???

@bigs @vyzo

bigs:
yeah good call
- **transactors** - accept writes, maintain journal
- **clients** - accept syndicated values from transactors, are primary query interface for reads
- **canonical** - permanent identifier, points to root

parkan:
**clients** could also read too, though, right? so read write or both, in theory

feel like there should be some distinct terms

bigs:
well a **client** can read head from a transactor but don't support writes unless they open up a session to **transactors** specifically, which is outside of spec

parkan:
@bigs do you mean supporting writes is outside the spec, or NOT supporting writes is? having trouble parsing that statement

my impression from our discussions was that the reader and writer client parts (which read HEAD/receive `journal-commit` events and send `put` commands, respectively) are logically somewhat separate

bigs:
ah... transactors support writes, but clients do not by their specification. ultimately, a client is just an interface for receiving a stream of updates to the chain. programmers implementing clients may choose to accept and propagate writes on behalf of users

parkan:
right, ok so then we need a different word for writers

bigs:
@parkan client-clients :stuck_out_tongue: 

edit: or plebians

parkan:
petitioners?

bigs:
maybe writers / readers / transactors / clients

parkan:
**folding:** producing a single flattened representation of a chain of updates, reconciling any conflicts and discarding bad information in the process

parkan:
Added here: https://github.com/mediachain/mediachain/wiki/Glossary

PR #109, opened 2016-06-01
RFC-3

autoencoder:
Putting up a draft for discussion. Don't have to merge (yet.)
- Is the glossary correct?
- Should we move the glossary to a separate document?

### DIFF
--- a/rfc/mediachain-rfc-3.md
+++ b/rfc/mediachain-rfc-3.md
@@ -14,7 +14,7 @@ We will also outline a roadmap toward improved, `Future` Indexer generations. Fu
 
 ## Terminology
 
-Relevant Mediachain terminology background for this RFC. See RFC-1[2] and RFC-2[3] for more information.
+Relevant Mediachain terminology background for this RFC. See also RFC-1[2] and RFC-2[3] for more information.
 
 - `Indexer` - Collection of Mediachain subsystems described in this RFC.
 - `Datastore` - Persistent content-addressed storage system, used to store Mediachain media and blocks. Currently IPFS[4].
@@ -22,7 +22,14 @@ Relevant Mediachain terminology background for this RFC. See RFC-1[2] and RFC-2[
 - `Folding` - Producing a single flattened representation of a chain of updates, reconciling any conflicts and discarding bad information in the process.
 - `Clients` - Embedded API or external nodes, which are the interface for end-user nodes to receive a stream of updates to the `Journal` from `Transactor` nodes, and for communicating writes to the `Transactors`. They also conduct reads and writes to the `Datastore`.
 - `Transactors` - Core of Mediachain. Accepts writes to the `Journal` from `Clients`, serves `Clients` with views of the `Journal`, and records transactions on their behalf. (Referred to as `Peer Nodes` in RFC-2.)
-- `Media Identifier` - One of several types of strings that can be used to indicate the identity of a media work. Most Indexer subsystems will resolve this identifier into a canonical ID. See more in the `Media Identifiers` section.
+- `Media Identifier` - String used to identify a media work. The string's prefix determines which, of several possible identifier types, each string is. See the `Media Identifiers` section in this RFC.
+- `Artefact` - Media work tracked by Mediachain. Identified in the blockchain by its latest `Canonical ID`.
+- `Entity` - Persona or organization who may associate with `Artefacts`. Identified in the blockchain by its latest `Canonical ID`.
+- `Canonical` -  Serialized blockchain block, stored in IPFS and identified by a `Canonical ID`.
+- `Canonical ID` - IPFS ID of a serialized blockchain block. In addition to identifying blockchain blocks, also used by blockchain to identify `Entities` and `Artifacts`. The current `Canonical IDs` of an `Entity` or `Artefact` can change over time, when its metadata is updated in the blockchain or when it is merged with another `Entity` or `Artefact`.
+- `Chain Head` - `Canonical ID` of latest blockchain block committed to the `Journal`. Multiple `Chain Heads` are permitted to exist during temporary blockchain forks.
+- `CanonicalEntry` - Type of `Journal` transaction that references `Canonical IDs` of `Entities` or `Artefacts`. Used to suggest that `Artefact` or `Entity` `Canonicals` should be added to the blockchain.
+- `ChainEntry` - Type of `Journal` transaction that references one or more `CanonicalEntry`. Used to suggest updates to latest state of the the blockchain.
 
 
 ## Roadmap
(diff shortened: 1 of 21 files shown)


rfc/mediachain-rfc-3.md
```diff
@@ -14,15 +14,22 @@ We will also outline a roadmap toward improved, `Future` Indexer generations. Fu
...
-- `Media Identifier` - One of several types of strings that can be used to indicate the identity of a media work. Most Indexer subsystems will resolve this identifier into a canonical ID. See more in the `Media Identifiers` section.
+- `Media Identifier` - String used to identify a media work. The string's prefix determines which, of several possible identifier types, each string is. See the `Media Identifiers` section in this RFC.
+- `Artefact` - Media work tracked by Mediachain. Identified in the blockchain by its latest `Canonical ID`.
+- `Entity` - Persona or organization who may associate with `Artefacts`. Identified in the blockchain by its latest `Canonical ID`.
+- `Canonical` -  Serialized blockchain block, stored in IPFS and identified by a `Canonical ID`.
+- `Canonical ID` - IPFS ID of a serialized blockchain block. In addition to identifying blockchain blocks, also used by blockchain to identify `Entities` and `Artifacts`. The current `Canonical IDs` of an `Entity` or `Artefact` can change over time, when its metadata is updated in the blockchain or when it is merged with another `Entity` or `Artefact`.
```
parkan: I don't think the Canonical and Canonical ID definitions are right--we've usually spoken of a "canonical" as the UUID-like identifier given to our "idea" of a work of art to which all data about it and its "instances" (copies, appearances in media, etc) refer to, and the "canonical id" as that identifier. In the current version, I **think** the correct analogy is the `Artefact`/`Entity` multihash (which is pointed to by `ref` in an `ArtefactCreationCell` etc) but we should clear this up.

parkan: the way that you're using this elsewhere it seems like maybe you want just "multihash" or "IPFS address"

rfc/mediachain-rfc-3.md
```diff
@@ -14,15 +14,22 @@ We will also outline a roadmap toward improved, `Future` Indexer generations. Fu
...
 - `Journal` - Flattened and partially-reconciled view of the Mediachain knowledge base. Maintained by `Transactors`.
 - `Folding` - Producing a single flattened representation of a chain of updates, reconciling any conflicts and discarding bad information in the process.
 - `Clients` - Embedded API or external nodes, which are the interface for end-user nodes to receive a stream of updates to the `Journal` from `Transactor` nodes, and for communicating writes to the `Transactors`. They also conduct reads and writes to the `Datastore`.
 - `Transactors` - Core of Mediachain. Accepts writes to the `Journal` from `Clients`, serves `Clients` with views of the `Journal`, and records transactions on their behalf. (Referred to as `Peer Nodes` in RFC-2.)
-- `Media Identifier` - One of several types of strings that can be used to indicate the identity of a media work. Most Indexer subsystems will resolve this identifier into a canonical ID. See more in the `Media Identifiers` section.
+- `Media Identifier` - String used to identify a media work. The string's prefix determines which, of several possible identifier types, each string is. See the `Media Identifiers` section in this RFC.
```
parkan: Sorry, I guess I'm reading this out of order -- these are what we'd call "canonical ids" (at least the IPFS style `Qm...` one)

rfc/mediachain-rfc-3.md
```diff
@@ -14,15 +14,22 @@ We will also outline a roadmap toward improved, `Future` Indexer generations. Fu
...
+- `Artefact` - Media work tracked by Mediachain. Identified in the blockchain by its latest `Canonical ID`.
+- `Entity` - Persona or organization who may associate with `Artefacts`. Identified in the blockchain by its latest `Canonical ID`.
+- `Canonical` -  Serialized blockchain block, stored in IPFS and identified by a `Canonical ID`.
```
parkan: I think this reflects some confusion about the design: conceptually, there are individual chains for each "Canonical" (aka Artefact/Entity), and a Journal with two parts: a blockchain containing blocks of (~1000 each) updates of the Chain Heads of these chains, and an index that points to the current chain head for each one. This is a bit of an unorthodox design and keeps tripping people up, we can do a call tomorrow with @vyzo (please correct me if my explanation here is off, too) to clear this up.

rfc/mediachain-rfc-3.md
```diff
@@ -14,15 +14,22 @@ We will also outline a roadmap toward improved, `Future` Indexer generations. Fu
...
+- `Canonical ID` - IPFS ID of a serialized blockchain block. In addition to identifying blockchain blocks, also used by blockchain to identify `Entities` and `Artifacts`. The current `Canonical IDs` of an `Entity` or `Artefact` can change over time, when its metadata is updated in the blockchain or when it is merged with another `Entity` or `Artefact`.
+- `Chain Head` - `Canonical ID` of latest blockchain block committed to the `Journal`. Multiple `Chain Heads` are permitted to exist during temporary blockchain forks.
```
parkan: This would be something like "the `head` field of the latest ChainEntry commited to the Journal for a given Canonical, referring to the latest ChainCell for that Canonical"

rfc/mediachain-rfc-3.md
parkan: (so we should sort out the terminology around this, particularly the ID)

parkan:
Left a bunch of specific comments, but the main takeaways are:
- the many chains + journal blockchain + journal index design isn't super clear from existing docs, we should 1) sync up the team to make sure we are on the same page now 2) create a doc (maybe not RFC format, something visual?) explaining this core design in simple terms
- need to clear up meaning and nature of IDs
- the glossary should probably be separate from the RFC, see also https://github.com/mediachain/mediachain/issues/29

rfc/mediachain-rfc-3.md
autoencoder: So the idea with the `Media Identifiers` is to have a name for all types of image identifiers which can be accepted by the Indexer APIs. In other words, `canonical IDs` are a subset of the possible identifier types the Indexer API can accept. Also currently supported is querying by image content e.g. data URIs of small images, or could even extend it to querying by image descriptors / vectors that represent abstract visual concepts instead of a particular image's content - as is sometimes done in some image QA systems.

rfc/mediachain-rfc-3.md
autoencoder: It'll be great to clarify how "idea" versus "instance" are each tracked in the system.

While we're at it, worth keeping in mind exactly how we'll define "idea" -- e.g. what's the particular definition of a sameness. The supervised indexer models will support being separately trained on multiple, potentially conflicting, definitions of sameness. At a minimum, 2 types - duplicate similarity and search query relevance similarity. But there can be multiple conflicting sub-types to choose from, for each of those as well.

At some point, also worth thinking about whether the chain or end-user applications might support tracking multiple conflicting sameness definitions, possibly for applying to different purposes.

rfc/mediachain-rfc-3.md
parkan: Yeah, totally agreed that all of these are valid "identifiers" but I think we should pick something as first class, which is the canonical/IPFS address in this case. Conversely, I don't think IPFS addresses of other IPFS objects should be considered "canonical ids". So all canonical ids are IPFS addresses, but not all IPFS addresses are canonical ids.

PR #110, opened 2016-06-01
RPC client interface for transactor

yusefnapora:
Still a work in progress.

So far, adds three RPC methods
- `FetchObjectChainHead`
  - takes a multihash reference and gets the head of a canonical's chain
- `InsertCanonicalRecord`
  - accepts a cbor-encoded Entity or Artefact and tries to insert
- `UpdateCanonicalRecord`
  - accepts a cbor-encoded `ChainCell` and tries to update the referenced canonical

It has these message types defined:
- `MultihashReference` - contains a base58-encoded string repr of a multihash
- `InsertRequest` - contains a byte array with cbor-encoded `Entity` or `Artefact`
- `UpdateRequest` - contains a byte array with cbor-encoded `ChainCell`
  - `FIXME`: this should be changed to include a ref to the canonical, to match the interface in the copycat client and let the transactor verify that the reference in the cell matches the one in the request.

Adds a simple scala cli app to start the RPC service, which wraps the copycat client interface.  To run with sbt:

``` shell
sbt 'transactor/run-main io.mediachain.transactor.RpcService <transactor-host:port> <rpc-service-port>
```

That will start a service that connects to the cluster at the given address and listens on the given port for rpc requests.

### DIFF
--- a/protocol/src/main/protobuf/Services.proto
+++ b/protocol/src/main/protobuf/Services.proto
@@ -51,3 +51,5 @@ service LSpaceService {
     rpc FetchCanonicalHistory(FetchCanonicalRequest) returns (CanonicalWithHistory);
     rpc ListWorksForAuthor(WorksForAuthorRequest) returns (WorksForAuthor);
 }
+
+
--- a/protocol/src/main/protobuf/Services.proto
+++ b/protocol/src/main/protobuf/Services.proto
@@ -51,4 +51,3 @@ service LSpaceService {
     rpc FetchCanonicalHistory(FetchCanonicalRequest) returns (CanonicalWithHistory);
     rpc ListWorksForAuthor(WorksForAuthorRequest) returns (WorksForAuthor);
 }
-
--- a/protocol/src/main/protobuf/Services.proto
+++ b/protocol/src/main/protobuf/Services.proto
@@ -52,4 +52,3 @@ service LSpaceService {
     rpc ListWorksForAuthor(WorksForAuthorRequest) returns (WorksForAuthor);
 }
 
-
--- a/protocol/src/main/protobuf/Transactor.proto
+++ b/protocol/src/main/protobuf/Transactor.proto
@@ -1,0 +1,16 @@
+syntax = "proto3";
+package protocol;
+option java_package = "io.mediachain.protocol.transactor";
+
+message FetchObjectChainRequest {
+    string objectID = 1;
+}
+
+message ObjectCell {
+    string objectBytes = 1;
+}
+
+service TransactorService {
+    rpc FetchObjectChainHead(FetchObjectChainRequest) returns
+    (stream ObjectCell);
+}
--- a/protocol/src/main/protobuf/Transactor.proto
+++ b/protocol/src/main/protobuf/Transactor.proto
@@ -2,15 +2,12 @@ syntax = "proto3";
 package protocol;
 option java_package = "io.mediachain.protocol.transactor";
 
-message FetchObjectChainRequest {
-    string objectID = 1;
-}
-
-message ObjectCell {
-    string objectBytes = 1;
+message MultihashReference {
+    string reference = 1;
 }
 
 service TransactorService {
-    rpc FetchObjectChainHead(FetchObjectChainRequest) returns
-    (stream ObjectCell);
+    rpc FetchObjectChainHead(MultihashReference) returns
+        (MultihashReference);
 }
+
--- a/protocol/src/main/protobuf/Transactor.proto
+++ b/protocol/src/main/protobuf/Transactor.proto
@@ -6,8 +6,22 @@ message MultihashReference {
     string reference = 1;
 }
 
+message InsertRequest {
+    bytes canonicalCbor = 1;
+}
+
+message UpdateRequest {
+    bytes chainCellCbor = 1;
+}
+
 service TransactorService {
     rpc FetchObjectChainHead(MultihashReference) returns
+        (MultihashReference);
+
+    rpc InsertCanonicalRecord(InsertRequest) returns
+        (MultihashReference);
+
+    rpc UpdateCanonicalRecord(UpdateRequest) returns
         (MultihashReference);
 }
 
--- a/protocol/src/main/protobuf/Transactor.proto
+++ b/protocol/src/main/protobuf/Transactor.proto
@@ -15,13 +15,13 @@ message UpdateRequest {
 }
 
 service TransactorService {
-    rpc FetchObjectChainHead(MultihashReference) returns
+    rpc InsertCanonical(InsertRequest) returns
         (MultihashReference);
 
-    rpc InsertCanonicalRecord(InsertRequest) returns
+    rpc UpdateChain(UpdateRequest) returns
         (MultihashReference);
 
-    rpc UpdateCanonicalRecord(UpdateRequest) returns
+    rpc LookupChain(MultihashReference) returns
         (MultihashReference);
 }
 
--- a/transactor/src/main/scala/io/mediachain/copycat/StateMachine.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/StateMachine.scala
@@ -15,7 +15,6 @@ import cats.data.Xor
 object StateMachine {
   import io.mediachain.protocol.Datastore._
   import io.mediachain.protocol.Transactor._
-  import io.grpc.{Server, ServerBuilder}
 
   val JournalBlockSize: Int = 4096 // blocksize for Journal Blockchain
   
@@ -126,12 +125,18 @@ object StateMachine {
     
     def lookup(commit: Commit[JournalLookup]): Option[Reference] = {
       try {
-        state.index.get(commit.operation.ref).flatMap(_.chain)
+        lookup(commit.operation.ref)
       } finally {
         commit.release()
       }
     }
-
+    
+    def lookup(ref: Reference): Option[Reference] = {
+      state.synchronized {
+        state.index.get(ref).flatMap(_.chain)
+      }
+    }
+    
     def currentBlock(commit: Commit[JournalCurrentBlock]) : JournalBlock = {
       try {
         JournalBlock(state.seqno, state.blockchain, state.block.toArray)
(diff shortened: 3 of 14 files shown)


transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  override def insertCanonicalRecord(request: InsertRequest)
+  : Future[Transactor.MultihashReference] = {
+    val recXor = CborSerialization.fromCborBytes[CanonicalRecord](
+      request.canonicalCbor.toByteArray
+    )
+
+    val future = recXor match {
+      case Xor.Left(err) => throw new StatusRuntimeException(
+        Status.FAILED_PRECONDITION.withDescription(
```
parkan: This should be an `INVALID_ARGUMENT`

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -0,0 +1,132 @@
...
+      )
+  }
+
+  override def insertCanonicalRecord(request: InsertRequest)
+  : Future[Transactor.MultihashReference] = {
+    val recXor = CborSerialization.fromCborBytes[CanonicalRecord](
+      request.canonicalCbor.toByteArray
+    )
+
+    val future = recXor match {
```
parkan: would prefer more descriptive `insertF` type naming rather than just `future`

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  override def updateCanonicalRecord(request: UpdateRequest)
+  : Future[Transactor.MultihashReference] = {
+    val cellXor = CborSerialization.fromCborBytes[ChainCell](
+      request.chainCellCbor.toByteArray
+    )
+
+    val future = cellXor match {
+      case Xor.Left(err) => throw new StatusRuntimeException(
+        Status.FAILED_PRECONDITION.withDescription(
```
parkan: `INVALID_ARGUMENT`

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -0,0 +1,132 @@
...
+                        (implicit executionContext: ExecutionContext)
+  : Unit = {
+    import scala.language.existentials
+
+    val builder = ServerBuilder.forPort(port)
+    val server = builder.addService(
+      TransactorServiceGrpc.bindService(service, executionContext)
+    ).build
+
+    executor.submit(new Runnable {
```
parkan, line 135: So the idea here is that there's a "master" gRPC thread (on `executor`) and a worker pool provided by the `executionContext` (which is by default a ForkJoinPool), right? Can you add some comments, either here or at the call?

parkan:
Whoa this crashed scalac, rebuilding

yusefnapora:
dang... we can revert the grpc version bump if it's a problem... We don't need anything specific from it, I just figured it'd be a good idea since we were a few point releases behind

parkan:
I think it was just a fluke

protocol/src/main/protobuf/Transactor.proto
```diff
@@ -0,0 +1,27 @@
...
+message InsertRequest {
+    bytes canonicalCbor = 1;
+}
+
+message UpdateRequest {
+    bytes chainCellCbor = 1;
+}
+
+service TransactorService {
+    rpc FetchObjectChainHead(MultihashReference) returns
```
vyzo: that's one ugly method name...

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -0,0 +1,132 @@
...
+    )
+
+    val insertF = recXor match {
+      case Xor.Left(err) => throw new StatusRuntimeException(
+        Status.INVALID_ARGUMENT.withDescription(
+          s"Object deserialization error: ${err.message}"
+        )
+      )
+      case Xor.Right(obj) =>
+        client.insert(obj)
```
vyzo, line 65: Can we do some sanity checking in the records before inserting to the journal?
At the very least we should check object size (let's say we limit to 64k initially) and that the meta is not empty.

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -0,0 +1,132 @@
...
+    )
+
+    val updateF = cellXor match {
+      case Xor.Left(err) => throw new StatusRuntimeException(
+        Status.INVALID_ARGUMENT.withDescription(
+          s"Object deserialization error: ${err.message}"
+        )
+      )
+      case Xor.Right(cell) =>
+        client.update(cell.ref, cell)
```
vyzo, line 96: here too

transactor/src/main/scala/io/mediachain/transactor/RpcService.scala
```diff
@@ -0,0 +1,40 @@
...
+      print("arguments: server-address rpc-service-port")
+      sys.exit(1)
+    }
+
+    val serverAddress = args(0)
+    val rpcPort = Integer.parseInt(args(1))
+
+    val client = Client.build()
+    logger.info(s"Connecting to server at $serverAddress...")
+    client.connect(serverAddress)
```
vyzo: we should add support for cluster connect now that the Client supports it and we have an actual cluster.

protocol/src/main/protobuf/Transactor.proto
```diff
@@ -0,0 +1,27 @@
...
+        (MultihashReference);
+
+    rpc InsertCanonicalRecord(InsertRequest) returns
+        (MultihashReference);
+
+    rpc UpdateCanonicalRecord(UpdateRequest) returns
```
vyzo: this method name is not exactly right -- we are not updating the canonical record, just its chain pointer.

transactor/src/main/scala/io/mediachain/copycat/StateMachine.scala
```diff
@@ -131,6 +131,7 @@ object StateMachine {
         commit.release()
       }
     }
+
```
vyzo: trim this, why diff on StateMachine?

protocol/src/main/protobuf/Services.proto
```diff
@@ -51,3 +51,5 @@ service LSpaceService {
     rpc FetchCanonicalHistory(FetchCanonicalRequest) returns (CanonicalWithHistory);
     rpc ListWorksForAuthor(WorksForAuthorRequest) returns (WorksForAuthor);
 }
+
```
vyzo: trim whitespace please

protocol/src/main/protobuf/Transactor.proto
vyzo: it is also inconsistent with the other two, what's the difference between an Object and a CanonicalRecord?

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo, line 135: I second this, it is not clear what thread is doing what here.

vyzo, line 135: I think we should make a fresh execution context instead of the global one for the service.

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -0,0 +1,132 @@
...
+  : Unit = {
+    import scala.language.existentials
+
+    val builder = ServerBuilder.forPort(port)
+    val server = builder.addService(
+      TransactorServiceGrpc.bindService(service, executionContext)
+    ).build
+
+    executor.submit(new Runnable {
+      override def run(): Unit = {
```
vyzo: uhm can you write this as:

```
def run {
 ...
}
```

I don't see the need for all that type noise here.

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo, line 96: i think we could also do some deeper checking here, like verifying the various references in the various chain cell types.

protocol/src/main/protobuf/Transactor.proto
parkan: you're right, CanonicalRecord is the right superclass -- I almost wonder if we should print out a big class hierarchy poster

parkan: I'm OK with kinda Java-y but descriptive method names overall tho

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
parkan, line 65: I think overall sanity checking is a separate discussion but object size check is probably good because we're writing to datastore directly from client

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
parkan, line 96: let's defer sanity checking discussion for the moment

protocol/src/main/protobuf/Transactor.proto
vyzo: I'll rename to 
- insertCanonical
- updateChain
- lookupChain
  short and descriptive enough.

protocol/src/main/protobuf/Transactor.proto
parkan: All our updates are append-only so this is equivalent, tho?

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo, line 65: I'll add max size checking -- let's start with 64k for now? We can easily scale up to 384K (near the dynamo max size) if need be.

protocol/src/main/protobuf/Transactor.proto
parkan: my issue with "updateChain/lookupChain" is that people seem to get the canonical chains and the blockchain confused (see my comments here https://github.com/mediachain/mediachain/pull/109#discussion_r65455130)

vyzo: The prime chain in user context is the mediachain for an entity or an artefact.
The user of the transactor does not care about the blockchain in this interface so there shouldn't be
an confusion.

PR #111, opened 2016-06-01
Serialize copycat commands as CBOR bytes

yusefnapora:
This adds custom copycat `TypeSerializer`s for the `JournalInsert`, `JournalUpdate` and `JournalLookup` commands, to fix the serialization bug when sending nested cbor maps in the `meta` field of an object.


vyzo:
I think we can do a little bit of common code refactoring -- reading and writing bytes at the minimum.

PR #112, opened 2016-06-02
scripts/ubuntu-setup fixes

vyzo:
git needs to be installed before we clone the repo and oracle license needs to be accepted (with debconf)


vyzo:
merging this, as it is verified to work.

PR #114, opened 2016-06-02
transactor: signal JournalDuplicateError on duplicate inserts

vyzo:
Facilitate deduplication with a specific error that includes the reference to the duplicate object.


yusefnapora:
awesome 👍

parkan:
:+1:

PR #115, opened 2016-06-03
Tweaks in transactor server

vyzo:
- Log the queue backlog when backing off in dynamo to help tune dynamo provisioning
- Add a simple interactive control loop for screened JournalServers, so that we can cleanly shutdown or retire a server

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
@@ -20,10 +20,10 @@ object JournalServer {
   }
   
   def parseArgs(args: Array[String]) = {
-    args match {
-      case Array("-i", config, cluster@_*) =>
+    args.toList match {
+      case "-i" :: config :: cluster =>
         (true, config, cluster.toList)
-      case Array(config, cluster@_*) =>
+      case config :: cluster =>
         (false, config, cluster.toList)
       case _ =>
         throw new RuntimeException("Expected arguments: [-i] config [cluster-address ...]")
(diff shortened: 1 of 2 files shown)


transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
```diff
@@ -4,26 +4,33 @@ import java.util.Properties
...
     val props = new Properties
-    props.load(new FileInputStream(propfile))
-    run(props, cluster)
+    props.load(new FileInputStream(config))
+    run(interactive, props, cluster)
   }
   
-  def run(conf: Properties, cluster: List[String]) {
+  def parseArgs(args: Array[String]) = {
+    args match {
```
parkan: given that you're `toList`ing inside anyway, why not

``` scala
args.toList match {
  "-i" :: config :: cluster =>  (true, config, cluster)
```

def more idiomatic scala than binding inside array extractors

transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
```diff
@@ -50,11 +57,45 @@ object JournalServer {
...
+      serverLoop(server)
+    }
+    datastore.close()
+  }
+  
+  def serverLoop(server: CopycatServer) {
     while (server.isRunning) {Thread.sleep(1000)}
   }
+
+  def interactiveLoop(server: CopycatServer) {
```
parkan, line 78: I would have probably just trapped SIGINT = shutdown and SIGHUP = leave but this is fine

transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
vyzo: ok, will fix

transactor/src/main/scala/io/mediachain/transactor/JournalServer.scala
vyzo, line 78: signal trapping and java don't play nicely -- i haven't found an effective way to trap signals.
this allows us to run on screen but also do a shell front-end that traps and feeds the right command.

PR #117, opened 2016-06-03
wip - streaming api for RPC service

yusefnapora:
So far untested and missing some important features (like catchup of previous blocks / events).  Compiles tho :)


protocol/src/main/protobuf/Transactor.proto
```diff
@@ -14,14 +14,34 @@ message UpdateRequest {
...
+    }
+}
+
 service TransactorService {
     rpc InsertCanonical(InsertRequest) returns
         (MultihashReference);
 
     rpc UpdateChain(UpdateRequest) returns
-        (MultihashReference);
+        (ChainUpdateDescription);
```
parkan: oh yeah 😅

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -3,21 +3,30 @@ package io.mediachain.copycat
...
+import io.grpc.stub.StreamObserver
 import io.grpc.{ServerBuilder, Status, StatusRuntimeException}
+import io.mediachain.copycat.Client.{ClientState, ClientStateListener}
 import io.mediachain.multihash.MultiHash
 import io.mediachain.protocol.CborSerialization
 import io.mediachain.protocol.Datastore._
-import io.mediachain.protocol.Transactor.JournalError
-import io.mediachain.protocol.transactor.Transactor.{InsertRequest, TransactorServiceGrpc, UpdateRequest}
+import io.mediachain.protocol.Transactor.{JournalError, JournalListener}
+import io.mediachain.protocol.transactor.Transactor.{MultihashReference => _, _}
```
parkan, line 15: are we trying to avoid shadowing another MultihashReference? also yusef you keep discovering new and strange ways of aliasing imports, good job

protocol/src/main/protobuf/Transactor.proto
parkan: I'm not sure I love the "switch two words around and add either Request or Description (and maybe drop a word)" method of deriving these names

parkan: I kind of want to just say `UpdateChainResult`

protocol/src/main/protobuf/Transactor.proto
```diff
@@ -14,14 +14,34 @@ message UpdateRequest {
...
 
     rpc LookupChain(MultihashReference) returns
         (MultihashReference);
+
+    rpc StreamJournalUpdates(JournalStreamRequest) returns
```
parkan: `rpc JournalStream(JournalStreamRequest)` ?

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -104,19 +113,92 @@ class TransactorService(client: Client,
...
+
+  override def streamJournalUpdates(request: JournalStreamRequest,
+    responseObserver: StreamObserver[JournalEvent]): Unit = {
+
+    // TODO: play event stream following lastBlockRef in request.
+    // this involves accessing the datastore, keeping local state
+    // per-observer, etc.
+
+    journalEventObservers.synchronized {
+      journalEventObservers.add(responseObserver)
```
parkan: paging @bigs since I think he researched this

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -104,19 +113,92 @@ class TransactorService(client: Client,
...
     }
   }
+
+
+  override def onStateChange(state: ClientState): Unit = {
+    logger.info(s"copycat state changed to $state")
+    if (state == ClientState.Disconnected) {
+      journalEventObservers.synchronized {
+        journalEventObservers.foreach { observer =>
+          observer.onError(
```
parkan: why did they call this `onError`? it's not a listener! 😞

protocol/src/main/protobuf/Transactor.proto
vyzo: why are we changing this return type?

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
yusefnapora, line 15: I don't know how that happened, lol 😛  Maybe intellij has become self aware.

protocol/src/main/protobuf/Transactor.proto
yusefnapora: yeah, this name kind of sucks.  I changed the return type to more closely match the ChainEntry, where you have a ref to the canonical and the cell instead of just the cell.  I figure it could be useful in a client, because you might only care about updates to a subset of all canonicals, so having the canonical ref could be handy.  `UpdateChainResult` sounds good tho

vyzo: The client already has the canonical in order to do the update, so it doesn't need it.
What could be perhaps useful is the chainPrevious in the ChainEntry.

protocol/src/main/protobuf/Transactor.proto
```diff
@@ -14,6 +14,24 @@ message UpdateRequest {
...
 }
 
+message JournalStreamRequest {
+    MultihashReference lastJournalBlock = 1;
+}
+
+message UpdateChainResult {
+    MultihashReference canonicalRef = 1;
+    MultihashReference chainHeadRef = 2;
+    MultihashReference chainPreviousHeadRef = 3;
```
vyzo: is there anything gained by calling things XXXRef instead of just XXX?

yusefnapora: no, I think you're right.  using Ref is inconsistent with the other protos as well

yusefnapora: How do you feel about `canonical`, `chain`, and `chainPrevious`?  It's closer to the ChainEntry naming but I think using `canonical` instead of `ref` is a bit clearer.   Maybe it's better to be consistent with ChainEntry though.

vyzo: that sounds much better.

transactor/src/main/scala/io/mediachain/transactor/Entrypoints.scala
```diff
@@ -0,0 +1,40 @@
...
+  def startCopycatService(listenAddr: String,
+                          logDir: String = "tmp"): Unit = {
+    import io.mediachain.copycat.Server
+    import io.mediachain.datastore.DynamoDatastore
+    import com.amazonaws.auth.BasicAWSCredentials
+
+    val datastoreConfig = DynamoDatastore.Config(
+      "Mediachain",
+      new BasicAWSCredentials("", ""),
+      Some("http://localhost:8000")
```
vyzo: why is this useful at all?
The servers are running in the cluster with a server program, you can already start a server by running JournalServer and the transactor server by running RpcService.
Also, it's hardcoded to local dynamo and uses a table we are not currently using (we are still on test).
Also the db name in dynamo comes from your access key, and there is none here which is bad.

vyzo:
@bigs the EntryPoints thing is useless I think. 
We arelady have a program to run the server, no need to have a hardcoded mess in the source tree. 
Please kill it!

protocol/src/main/protobuf/Transactor.proto
```diff
@@ -14,6 +14,24 @@ message UpdateRequest {
...
+    MultihashReference chain = 2;
+    MultihashReference chainPrevious = 3;
+}
+
+message JournalEvent {
+    oneof event {
+        MultihashReference canonicalInserted = 2;
+        UpdateChainResult chainUpdated = 3;
+        MultihashReference journalBlockPublished = 4;
+    }
```
vyzo, line 32: naming-wise, can we drop the past tense from the events?

vyzo:
belay the comment about toStream -- I just read the diff and thought it was in Datastore!

protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
```diff
@@ -354,5 +359,19 @@ object Datastore {
       val optionals = Map("chain" -> chain.map(_.toCbor))
       super.toCMapWithDefaults(defaults, optionals)
     }
+
+    def toStream(datastore: Datastore): Streaming[JournalBlock] = {
+      val next: Eval[Streaming[JournalBlock]] = chain match {
+        case Some(ref) => Eval.always {
```
parkan, line 386: is the point of `Eval` here to avoid blowing the stack?

parkan:
@bigs let's address https://github.com/mediachain/mediachain/commit/c124ad791ec068245fedcb1f45094e4f51abb33b#commitcomment-17783823 and I'll merge this.

Also agree with https://github.com/mediachain/mediachain/pull/117#discussion_r66030131 but that can go on top

protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
bigs, line 386: `Eval.always` means "always execute this block" instead of caching the result

parkan, line 386: right, I get what `Eval` generally does, but given that the two branches here are `.always` and a `.now` returning a constant, I'm assuming it's used as a trampoline for the recursion.

parkan:
@bigs let's address https://github.com/mediachain/mediachain/commit/c124ad791ec068245fedcb1f45094e4f51abb33b#commitcomment-17783823 and merge?

bigs:
@parkan that entire file can be removed from the project. it's been superseded by the main function yusef and i wrote

yusefnapora:
I thought @bigs meant to remove the `Entrypoints`, not the `RpcService`.   I think we still need `RpcService` to run the facade.  But `Entrypoints` is just for easier repl development.

parkan:
Oops yeah I misread that one sec

parkan:
Actually both RpcService and Entrypoints hardcode the DynamoDatastore as local, what's the deal? @bigs

Issue #116, opened 2016-06-03
Mediachain backup in S3


vyzo:
In order to protect from disasters we should back up the mediachain in S3.
This can follow block generation: every time a new block is generated, a supporting node would aggregate the block together with all the new objects pointed by the block together into a fat block and store it in S3.

This allows us to recover from disasters while losing at most one block of updates.
It also simplifies adding IPFS pinning nodes that provide the whole mediachain, as such nodes can use the fat blocks directly from S3.

denisnazarov:
@vyzo Not an immediate concern, but how do we guarantee trust here, eg. that Mediachain hasn't tampered with the contents in S3? Periodic timestamp in the bitcoin blockchain?

vyzo:
We can do that.

parkan:
yeah there's plenty of ways, we can just write the latest block height + hash to any number of proof of existence services

vyzo:
Publishing the hash of the block should be sufficient, as all the other objects are hash-pointed.

parkan:
true, just nice to have block height for reference purposes too

parkan:
so you can have like https://blockchain.info/block/000000000000000000c639b066db9b266c64b5787840d9b1f7d7662bb1c286db

vyzo:
oh yeah, the index can be there just fine; I was clarifying the sufficiency of hash publishing.

vyzo:
Implemented with #118 and #122 
The restore is a little slow as it has to replay the blockchain in order to be idempotent, which limits throughput.

PR #118, opened 2016-06-07
Make original metadata reference in records a first-class citizen

vyzo:
Adds a metaSource field to the Record data model so that they can be tracked by the mediachain as first class objects. 
#116 This allows us to back them up together with the block with a backing store client.


yusefnapora:
Looks good 👍  

Only thing that might be nice to add is the sanity checking you were suggesting yesterday (that the metaSource exists), but we can do that in a separate pass.

PR #120, opened 2016-06-08
Fix the wrong url link in rfc md files.

genggjh:
Signed-off-by: David Geng david@esse.io


parkan:
Good catch, thank you

PR #122, opened 2016-06-08
S3 backing store

### LINKED ISSUE #116: Mediachain backup in S3

vyzo:
close #116 
Adds support for backing up and restoring the mediachain in S3:
- S3BackingStore implements the backup process as a transactor client which archives fat blocks into S3
- S3Restore implements restore by replaying the blockchain from S3 backed up fat blocks.

### DIFF
--- a/protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
+++ b/protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
@@ -608,6 +608,8 @@ object CborSerialization {
 
   object JournalBlockArchiveDeserializer extends CborDeserializer[JournalBlockArchive]
   {
+    import scala.annotation.tailrec
+
     def fromCMap(cMap: CMap): Xor[DeserializationError, JournalBlockArchive] =
       for {
         _ <- assertRequiredTypeName(cMap, MediachainTypes.JournalBlockArchive)
@@ -624,7 +626,7 @@ object CborSerialization {
         )
     
     def dataFromCbor(records: CArray) = {
-      def loop(rest: List[CValue], map: Map[Reference, Array[Byte]])
+      @tailrec def loop(rest: List[CValue], map: Map[Reference, Array[Byte]])
       : Xor[DeserializationError, Map[Reference, Array[Byte]]] = {
         rest match {
           case CArray(List(refCbor: CMap, data: CBytes)) :: rest =>
--- a/protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
+++ b/protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
@@ -178,6 +178,7 @@ object CborSerialization {
         case CanonicalEntry.stringValue => Xor.right(CanonicalEntry)
         case ChainEntry.stringValue => Xor.right(ChainEntry)
         case JournalBlock.stringValue => Xor.right(JournalBlock)
+        case JournalBlockArchive.stringValue => Xor.right(JournalBlockArchive)
 
         case _ => Xor.left(UnexpectedObjectType(string))
       }
@@ -234,6 +235,9 @@ object CborSerialization {
     case object JournalBlock extends MediachainType {
       val stringValue = "journalBlock"
     }
+    case object JournalBlockArchive extends MediachainType {
+      val stringValue = "journalBlockArchive"
+    }
 
     val ArtefactChainCellTypes: Set[MediachainType] = Set(
       ArtefactChainCell, ArtefactUpdateCell, ArtefactLinkCell, ArtefactCreationCell,
@@ -268,7 +272,8 @@ object CborSerialization {
 
       MediachainTypes.CanonicalEntry -> CanonicalEntryDeserializer,
       MediachainTypes.ChainEntry -> ChainEntryDeserializer,
-      MediachainTypes.JournalBlock -> JournalBlockDeserializer
+      MediachainTypes.JournalBlock -> JournalBlockDeserializer,
+      MediachainTypes.JournalBlockArchive -> JournalBlockArchiveDeserializer
     )
 
   val datastoreDeserializers: DeserializerMap =
@@ -599,6 +604,45 @@ object CborSerialization {
           chain = getOptionalReference(cMap, "chain"),
           entries = entries
         )
+  }
+
+  object JournalBlockArchiveDeserializer extends CborDeserializer[JournalBlockArchive]
+  {
+    def fromCMap(cMap: CMap): Xor[DeserializationError, JournalBlockArchive] =
+      for {
+        _ <- assertRequiredTypeName(cMap, MediachainTypes.JournalBlockArchive)
+        ref <- getRequiredReference(cMap, "ref")
+        xblock <- getRequired[CMap](cMap, "block")
+        block <- JournalBlockDeserializer.fromCMap(xblock)
+        xdata <- getRequired[CArray](cMap, "data")
+        data <- dataFromCbor(xdata)
+      } yield
+        JournalBlockArchive(
+          ref = ref,
+          block = block,
+          data = data
+        )
+    
+    def dataFromCbor(records: CArray) = {
+      def loop(rest: List[CValue], map: Map[Reference, Array[Byte]])
+      : Xor[DeserializationError, Map[Reference, Array[Byte]]] = {
+        rest match {
+          case CArray(List(refCbor: CMap, data: CBytes)) :: rest =>
+            ReferenceDeserializer.fromCMap(refCbor) match {
+              case Xor.Right(ref) =>
+                loop(rest, map + (ref -> data.bytes))
+              case err@Xor.Left(_) => err
+            }
+            
+          case hd :: _ =>
+            Xor.Left(UnexpectedObjectType(s"Archive data decoding failed; unexpected object: $hd"))
+            
+          case Nil => Xor.Right(map)
+        }
+      }
+      
+      loop(records.items, Map())
+    }
   }
 
   object CanonicalEntryDeserializer extends CborDeserializer[CanonicalEntry]
--- a/transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
@@ -1,15 +1,13 @@
 package io.mediachain.copycat
 
 import java.util.Properties
-import java.util.concurrent.{ExecutorService, Executors, TimeUnit, LinkedBlockingQueue}
+import java.util.concurrent.{ExecutorService, Executors, TimeUnit}
 import java.io.{File, FileOutputStream}
-import java.nio.file.Files
 import com.amazonaws.AmazonClientException
 import com.amazonaws.auth.{AWSCredentials, BasicAWSCredentials}
 import com.amazonaws.services.s3.AmazonS3Client
 import org.slf4j.{Logger, LoggerFactory}
-import scala.concurrent.{Future, Await}
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Await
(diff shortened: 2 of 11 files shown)


protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
```diff
@@ -601,6 +606,45 @@ object CborSerialization {
...
+          block = block,
+          data = data
+        )
+    
+    def dataFromCbor(records: CArray) = {
+      def loop(rest: List[CValue], map: Map[Reference, Array[Byte]])
+      : Xor[DeserializationError, Map[Reference, Array[Byte]]] = {
+        rest match {
+          case CArray(List(refCbor: CMap, data: CBytes)) :: rest =>
+            ReferenceDeserializer.fromCMap(refCbor) match {
```
parkan, line 633: I feel like this would be easier to read as a flatMap, since you're not doing anything with the other branch and it'll short-circuit anyway

protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
```diff
@@ -601,6 +606,45 @@ object CborSerialization {
...
+              case Xor.Right(ref) =>
+                loop(rest, map + (ref -> data.bytes))
```
parkan, line 635: I'm kind of surprised you didn't `reverse` the list and build this through normal returns from the tail recursion

bigs, line 635: :+1:

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
```diff
@@ -0,0 +1,252 @@
...
+  client.addStateListener(this)
+
+  def connect(addrs: List[String]) {
+    cluster = Some(addrs)
+    client.connect(addrs)
+    client.listen(this)
+  }
+  
+  def close() {
+    exec.shutdownNow()
```
parkan, line 42: do we really need to aggressively shut down everything active?

protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
vyzo, line 633: flat mapping will break tail recursion, as it won't be able to do a goto from the closure inside flatMap.

protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
vyzo, line 635: It feels more natural to me to do a folding loop using an accumulator; kind of a scheme idiom.
Also, the order of iteration is not important, so the reverse doesn't add anything.

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
vyzo, line 42: It is kind of a safe operation -- a restart will pick up where you stopped and you can't have incomplete puts in S3.
Also,  if you are calling close here you are probably exiting the entire program and you won't it to stop asap.

vyzo, line 42: I'll change it to plain shutdown to let any actual writes in progress a chance to complete without an InterruptedException.

vyzo, line 42: Actually plain shutdown is not quite correct, because we proceed to close the client used by the background threads anyway. Better to immediately shutdown with interrupt.
Will revert the change.

parkan, line 42: hmm ok maybe let's revisit this later

protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
```diff
@@ -601,6 +606,47 @@ object CborSerialization {
...
+        data <- dataFromCbor(xdata)
+      } yield
+        JournalBlockArchive(
+          ref = ref,
+          block = block,
+          data = data
+        )
+    
+    def dataFromCbor(records: CArray) = {
+      @tailrec def loop(rest: List[CValue], map: Map[Reference, Array[Byte]])
```
parkan, line 629: hmm you're right using flatMap breaks the @tailrec, I'd have assumed that the two statements would get compiled to equivalent bytecode :\

parkan, line 629: my apologies to the user named "tailrec", they must get this a lot

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
```diff
@@ -0,0 +1,248 @@
...
+    logger.info("Fetching current block")
+    val op = client.currentBlock
+    val block = Await.result(op, Duration.Inf)
+    logger.info("Current blockchain pointer is " + block.chain)
+    block.chain.foreach { execWriteBlock(_) }
+  }
+  
+  private def execWriteBlock(ref: Reference) {
+    this.synchronized {
+      if (!blocks.contains(ref)) {
```
parkan, line 92: am I reading this wrong, or will this back up the _entire_ chain every time?

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
```diff
@@ -0,0 +1,248 @@
...
+  
+  // JournalListener
+  def onJournalCommit(entry: JournalEntry) {}
+
+  def onJournalBlock(ref: Reference) {
+    logger.info(s"New block ${ref}")
+    execWriteBlock(ref)
+  }
+  
+  private def withErrorLog(expr: => Unit) {
```
parkan, line 73: I was about to say "why don't we just collect the futures and attach an onFailure handler" but then I went to read about java Futures and got really sad.

I kind of want to just rewrite everything like

```
implicit val execContext = ExecutionContext.fromExecutor(exec)
def logError = PartialFunction[Exception, Unit] = { case e: Exception => ... }
Future { writeCurrentBlock() }.recover(logError)
...
```

(maybe `Future { blocking....`)

since s3 client doesn't even use java futures, right?

transactor/src/main/scala/io/mediachain/copycat/S3Restore.scala
```diff
@@ -0,0 +1,200 @@
...
+  import S3BackingStore.multihashRef
+  private val logger = LoggerFactory.getLogger(classOf[S3Restore])
+  private val datastore = new DynamoDatastore(config.dynamo)
+  private val s3 = new AmazonS3Client()
+  private val s3bucket = config.s3bucket
+  private val client = Client.build(config.sslConfig)
+  // dynamo background writing
+  @volatile private var shutdown = false
+  private val queue = new LinkedBlockingQueue[(Reference, Array[Byte])]
+  private val writers = (1 to Runtime.getRuntime.availableProcessors).map { n =>
```
parkan, line 31: decided to manually manage these instead of using a pool?

transactor/src/main/scala/io/mediachain/copycat/S3Restore.scala
```diff
@@ -0,0 +1,200 @@
...
+    logger.info(s"Restoring ${cluster} from ${key}")
+    client.connect(cluster)
+    doRestore(key)
+    close()
+  }
+  
+  def close() {
+    logger.info("Shutting down...")
+    shutdown = true
+    writers.foreach(_.join())
```
parkan, line 50: don't need to do this w/pool

parkan:
A lot of stuff happens here, for the most part it looks OK (other than huge scope for one PR!). Not super into the java-style error and thread handling but I get that it's kind of informed by using the s3 libs etc that don't even use futures of any kind.

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
vyzo, line 92: It will walk up the chain until it finds a block it already has in the list of visited blocks or reaches nil.
It will not write a block a second time, because it will either be in the visited blocks set (which will make execWriteBlock a noop) or already in S3 (the write checks to see if it is already present in S3 with listObjects)

The reason for this implementation is that it allows the backing store to catch up in the block chain and be easily restartable and tolerant to transient client failures.

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
vyzo, line 73: s3 interface is synchronous (as are all service interfaces in AWS sdk)

transactor/src/main/scala/io/mediachain/copycat/S3Restore.scala
vyzo, line 31: yeah, I find it cleaner when the threads are specialized.

vyzo:
Yeah, we kind of have to write synchronous java-style code when using the aws sdk.

vyzo:
I think we want to add backoff/retry around S3 errors as well - the one thing currently missing.

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
parkan, line 92: ah ok the listObjects call inside the write was what I was missing (concern was that the set wouldn't be populated on a repeated run)

transactor/src/main/scala/io/mediachain/copycat/S3Restore.scala
parkan, line 31: the "preferred" java way to do that is

```
namedThreadFactory = new ThreadFactoryBuilder()
                                       .setNameFormat("S3Restore@${this.hashCode}#write-%d").build()
```

and pass it to executor

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
parkan, line 73: right, so you would wrap them like this

``` scala
def putObject(s3bucket: String, key: String, gzpath: String): Future = Future {
  blocking {
    s3.putObject(s3bucket, key, new File(gzpath)
  }
}
```

this gives you a nice future-based interface and also some intelligent thread creation if you use ForkJoinPool

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
```diff
@@ -0,0 +1,255 @@
...
+    } catch {
+      case e: Throwable =>
+        logger.error("Unhandled exception in task", e)
+    }
+  }
+  
+  private def writeCurrentBlock() {
+    logger.info("Fetching current block")
+    val op = client.currentBlock
+    val block = Await.result(op, Duration.Inf)
```
parkan, line 85: is this Inf for a good reason?

parkan:
yeah agreed on backoff-retry

transactor/src/main/scala/io/mediachain/copycat/S3BackingStore.scala
vyzo, line 85: there is no reasonable timeout here -- and the logic would have to retry it anyway, with multiple pending currentBlocks.

transactor/src/main/scala/io/mediachain/copycat/S3Restore.scala
vyzo: ugh -- a builder for the factory!

PR #124, opened 2016-06-10
Provisioning

yusefnapora:
Adds a Vagrantfile and ansible config to install oracle java, sbt and dynamodb-local

Also adds an `io.mediachain.transactor.Main` object to act as a unified entry point for both the JournalServer and RpcService.  

check out this fancy usage text:

```
mediachain
Usage: mediachain [transactor|facade] [options]

  -l <host:port> | --listen <host:port>
        bind address
  -c <host1:port1>,<host2:port2> | --cluster-addresses <host1:port1>,<host2:port2>
        addresses of transactor cluster members
  -t <value> | --table-name <value>
        base name of dynamo-db table
  -s <value> | --ssl-properties <value>
        path to properties file for ssl configuration
  -e <value> | --dynamo-endpoint <value>
        endpoint for dynamo-db
Command: transactor [options]
run the transactor
  -d <value> | --data <value>
        directory to store transactor data
  -i | --interactive
        accept control commands from standard input
Command: facade
run an RPC facade to the transactor cluster
```

the ssl config still uses the property file, and the AWS access and secret keys are expected to come from the aws cli configuration.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,39 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure(2) do |config|
+  # Use a box compatible with both virtualbox and vmware_fusion providers
+  config.vm.box = "phusion/ubuntu-14.04-amd64"
+
+  # Forward transactor rpc service port 10001
+  config.vm.network "forwarded_port", guest: 10001, host: 10001
+  
+  # Forward port for local dynamodb (for development)
+  config.vm.network "forwarded_port", guest: 8000, host: 8000
+
+  # Forward jvm debugger port 5005 on the guest to 5006 on the host
+  config.vm.network "forwarded_port", guest: 5005, host: 5006
+
+  # Set ram, etc for vm providers
+  config.vm.provider "virtualbox" do |vb|
+    vb.memory = "2048"
+  end
+
+  config.vm.provider "vmware_fusion" do |vm|
+    vm.memory = "2048"
+  end
+
+  # Set hostname for vm
+  config.vm.define "transactor"
+
+  # Ansible config
+  config.vm.provision :ansible do |ansible|
+    ansible.limit = "transactors"
+    ansible.playbook = "meta/dev.yml"
+    ansible.groups = { "transactors" => ["transactor"] }
+  end
+end
--- a/meta/dev.yml
+++ b/meta/dev.yml
@@ -1,0 +1,6 @@
+---
+- hosts: transactors
+  roles:
+    - { role: jpnewman.java, tags: ["init"], become: yes, become_user: "root" }
+    - { role: milinda.sbt, become: yes, become_user: "root" }
+    - { role: dynamodb_local, become: yes, become_user: "root" }
--- a/meta/roles/dynamodb_local/handlers/main.yml
+++ b/meta/roles/dynamodb_local/handlers/main.yml
@@ -1,0 +1,4 @@
+- name: "Restart DynamoDB"
+  service: >
+    name=dynamodb
+    state=restarted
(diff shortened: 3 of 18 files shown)


meta/dev.yml
```diff
@@ -0,0 +1,6 @@
+---
+- hosts: transactors
+  roles:
+    - { role: jpnewman.java, tags: ["init"], become: yes, become_user: "root" }
+    - { role: milinda.sbt, become: yes, become_user: "root" }
+    - { role: dynamodb_local, become: yes, become_user: "root" }
```
vyzo, line 6: you reall want to run those as root? use a less priviledged user please, even if it's just for local testing.

Vagrantfile
```diff
@@ -0,0 +1,39 @@
+# -*- mode: ruby -*-
```
vyzo, line 1: Can we put the Vagrantfile somewhere deeper in the hierarchy?
It shouldn't be at the top of Mediachain tree as it is only for local debugging setup.

meta/roles/dynamodb_local/handlers/main.yml
```diff
@@ -0,0 +1,4 @@
+- name: "Restart DynamoDB"
+  service: >
+    name=dynamodb
+    state=restarted
```
vyzo, line 4: missing newline at end of file.

transactor/src/main/scala/io/mediachain/datastore/DynamoTableMaker.scala
```diff
@@ -0,0 +1,52 @@
+package io.mediachain.datastore
+
+object DynamoLocalTableMaker {
```
vyzo, line 3: Call this DynamoSetup or something please; this name is horrible!

transactor/src/main/scala/io/mediachain/datastore/PersistentDatastore.scala
```diff
@@ -94,7 +96,7 @@ class PersistentDatastore(config: PersistentDatastore.Config)
 
 object PersistentDatastore {
   case class Config(
-    dynamo: DynamoDatastore.Config, 
+    dynamo: DynamoConfig,
```
vyzo: Don't do this, I am also making changes in the config in my branch.

vyzo:
well great; you just created a horrible conflict with my branch by mudding the JournalServer and the configs as I am also working on them.

vyzo:
Also why are we moving away from property files for configuration?
I very much prefer (and it is a lot better from operational perspective and debugging) to set the configuration in a property file than have to fight with command line options.

transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
```diff
@@ -1,27 +1,28 @@
...
-  val creds = config.awscreds
   val table = config.baseTable
   val chunkTable = table + "Chunks"
   val chunkSize = 1024 * 384 // 384 KB; DynamoDB has 400KB limit for item size.
                              // this item size includes field names as well as
                              // values.
 
-  val db = new AmazonDynamoDBClient(creds)
-  config.endpoint.foreach(ep => db.setEndpoint(ep))
+  val db = new AmazonDynamoDBClient()
```
vyzo: and how do we pass credentials to the amazon db client?

vyzo:
And also, what happened to the aws creds?
Are we pulling these from env vars now?

vyzo:
General observations:
- move the Vagrantfile and meta stuff under a vagrant/ directory
- please add an option to the arguments for the servers to read configuration from a property file (and make command line options override the config)

vyzo:
Also, can you leave the definition of the dynamo config where it was originally?
It is fine to rename it, but it should be close to its usage, not inside a different package.
This will also make life easier in merging my changes.

meta/roles/dynamodb_local/meta/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  platforms:
+    - name: Ubuntu
+      versions:
+        - trusty
+  categories:
+    - development
+    - database
+    - database:nosql
+dependencies:
+  - role: jpnewman.java
```
vyzo: missing newline.

vyzo:
I would also recommend not changing the main function of JournalServer and leave the property parsing logic for standalone usage -- there is no reason to break existing clients/users.
You can invoke run directly from main with the correct options -- or add an overloaded run method that unpacks the new config object.

PR #126, opened 2016-06-13
Provisioning - 2nd try :)

yusefnapora:
Basically #124 with fewer breaking changes.  Based on master instead of the streaming dev branch.

Instead of changing JournalServer.run and removing JournalServer.main, overloads JournalServer.run with one that accepts a Config and translates that into the args expected by the existing run method.

Also, puts DynamoDatastore.Config back where it was, but removes the aws credentials, so we can use whatever has been configured using amazons tools.

Vagrant / ansible changes:

Moves `Vagrantfile` and the ansible `meta` directory into the new `provisioning` directory, instead of putting them in the project root.  To run vagrant, you should either set the `VAGRANT_CWD` environment variable to `$PWD/provisioning` or use `./scripts/vagrant.sh`, which will set the variable for you and pass any args on to the real `vagrant` command.   So, e.g. `./scripts/vagrant.sh up` will start the vm.

Note that, because the Vagrantfile is not at the project root, inside the vm `/vagrant` is mapped to the `provisioning` subdirectory.  The project root is available at `/mediachain` inside the vm.

Things I'd like to do:
- [ ] pull property file parsing into the Config parsing phase, so we can optionally use a property file instead of command line args for the "unified" entry point
- [ ] try to get private networking to work with my vagrant setup and stop forwarding ports to the host machine.
- [ ] possibly add an alternate multi-machine Vagrantfile to use to spin up a test cluster.  Would obviously need more resources, so shouldn't be the default :)


vyzo:
just needs a little whitespace trimming; :+1:

parkan:
@bigs could use your eye on this as resident "have used ansible before" expert

provisioning/meta/dev.yml
```diff
@@ -0,0 +1,6 @@
+---
```
bigs: i guess this should be in `playbooks`

imo we should choose, `provisioning` or `meta` as a tld. i use `meta` by convention, but don't really care too much. both seems... unnecessary :stuck_out_tongue:

provisioning/meta/playbooks/transactor-setup.yml
```diff
@@ -0,0 +1,20 @@
+---
+- hosts: transactor-provisioning
+  tasks:
+  - name: update system
+    apt: upgrade=yes
+  - name: setup git ppa repo
+    apt_repository: repo=ppa:git-core/ppa
+  - name: install git
```
bigs: why are we not using the `git` install present in the ubuntu repo

provisioning/meta/playbooks/transactor-setup.yml
```diff
@@ -0,0 +1,20 @@
...
+    apt: name=git state=present update_cache=true
+  - name: checkout mediachain
+    git: repo=https://github.com/mediachain/mediachain.git dest=/home/ubuntu/mediachain
+  - name: run setup script
+    shell: /home/ubuntu/mediachain/scripts/ubuntu-setup.sh && touch /home/ubuntu/.ubuntu-setup
```
bigs: let's adapt this to use ansible instead of being a shell script. ansible is idempotent and offers a lot of flexibility

provisioning/meta/playbooks/transactor-setup.yml
```diff
@@ -0,0 +1,20 @@
...
+    args:
+      creates: /home/ubuntu/.ubuntu-setup
+  - name: set ownership of mediachain src tree
+    file: path=/home/ubuntu/mediachain owner=ubuntu group=ubuntu recurse=yes
+  - name: prepare /mnt/transactor
```
bigs: are we using ebs?

provisioning/meta/roles/dynamodb_local/meta/main.yml
```diff
@@ -0,0 +1,16 @@
+galaxy_info:
+  author: Cole Brown <cole.brown@gmail.com>
+  description: Installs the development version of the Amazon DynamoDB server
```
bigs: we should drop a comment explaining we hijacked the one existing on ansible galaxy

yusefnapora: ah yeah, good point.

provisioning/meta/playbooks/transactor-setup.yml
vyzo: It's very old -- 1.9.x something.
I also were getting some 404s for the deb in repo in earlier instance setups, out of which this playbook evolved.

scripts/vagrant.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+type vagrant >/dev/null 2>&1 || { echo >&2 "Please install vagrant: https://www.vagrantup.com/"; exit 1; }
+
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PROJECT_ROOT="$SCRIPT_DIR/.."
```
bigs, line 7:
``` bash
PROJECT_ROOT=`git rev-parse --show-toplevel`
```

provisioning/meta/playbooks/transactor-setup.yml
vyzo: not for transactors, the data goes to ephemeral storage.

bigs:
generally fine w/ changes. few small things here and there. only major request is `ubuntu-setup.sh` becomes an ansible role

provisioning/meta/playbooks/transactor-setup.yml
vyzo: Let's do that the next time around? I am kind of content with state of affairs as it is right now.

scripts/vagrant.sh
yusefnapora, line 7: woah, cool

PR #136, opened 2016-06-16
Additional provision script for transactor clients

vyzo:
Like transactor, but also pulls medichain-client and configures /mnt/data instead of /mnt/transactor.

### DIFF
--- a/provisioning/playbooks/transactor-client-provisioning.yml
+++ b/provisioning/playbooks/transactor-client-provisioning.yml
@@ -4,6 +4,12 @@
     - transactor
 
   tasks:
+  - name: install pip
+    apt: name=python-pip state=present
+
+  - name: update to latest pip
+    shell: pip install --upgrade pip
+    
   - name: checkout mediachain
     git: repo=https://github.com/mediachain/mediachain.git dest=/home/ubuntu/mediachain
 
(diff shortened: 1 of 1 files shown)


provisioning/playbooks/transactor-client-provisioning.yml
```diff
@@ -0,0 +1,26 @@
+---
+- hosts: transactor-client-provisioning
+  roles:
+    - transactor
+
+  tasks:
+  - name: install pip
+    apt: name=python-pip state=present
+
+  - name: update to latest pip
```
parkan, line 10: so this internally runs `pip install -U pip`?

parkan, line 10: ah yeah I see, you're using `pip` to install `pip` to `latest`, makes sense

provisioning/playbooks/transactor-client-provisioning.yml
```diff
@@ -0,0 +1,26 @@
...
+    pip: name=pip state=latest
+    
+  - name: checkout mediachain
```
bigs, line 13: maybe add a variable `mediachain_branch` that defaults to `master` here

bigs:
i kno it's a bit of a pain (& would be happy to help) but it would nice to have this in a role so we can reuse it/compose it. looks good

vyzo:
I've been thinking of refactoring both provisioning scripts to extract some roles, so will do.

vyzo:
I think I'll leave the role refactoring for next pass; the pain level exceeds my current tolerance :)

PR #137, opened 2016-06-17
Server control with killfiles and common interfaces

vyzo:
- Add a ServerControl class that watches a directory for killfiles and dispatches commands.
- Updates JournalServer, RpcService, and S3BackingStoreClient to use ServerControl for orderly shutdown.
- Refactors properties usage in various server components and unifies their configuration.

### DIFF
--- a/transactor/README.md
+++ b/transactor/README.md
@@ -1,9 +1,28 @@
-# mediachain-transactor
+Transactors provide read/write access to the Mediachain Journal and syndication.
 
-Consensus/journal cluster server for the mediachain network. To run a test node:
+# How to Run a test transactor node
 
-* `brew install dynamodb-local && brew services start dynamodb-local`
-* Prep DynamoDB tables
+## Compile the transactor code
+```
+sbt transactor/assembly
+```
+This will leave a fat jar in transactor/target/scala-2.11/, which we call
+the `BIGJAR`
+
+## Local DynamoDB setup
+
+The current transactor implementation uses DynamoDB as the datastore in place
+of IPFS (at least until IPLD is rolled out).
+
+For testing you probably want to use the local implementation of DynamoDB.
+See [https://aws.amazon.com/blogs/aws/dynamodb-local-for-desktop-development/](DynamoDB Local)
+
+On a Mac you can install it with homebrew:
+```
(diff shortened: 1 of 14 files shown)


yusefnapora:
I'm fine with using properties for configuration, but I think we need to add a list of property keys and expected values somewhere (the README at least, but maybe also as a `--help` output for the cli parser?).  

What I liked about the old Config wasn't really the command line args per se, but more that all the possible config options were gathered in one place, and it was possible to find out what they were without digging into a bunch of places in the code.  But I think we can have that with property files if we just gather the keys somewhere and spit them out.

vyzo:
I will update the README for the properties in use.

yusefnapora:
Awesome.  👍

transactor/README.md
```diff
@@ -24,26 +43,50 @@ val chunkTableAttrs = List(new AttributeDefinition("chunkId", "S"))
...
 io.mediachain.transactor.ssl.keyStoreKeyPassword: $KEYSTOREKEYPASS
 ```
 
-* `sbt transactor/assembly`
-* `scala -cp $BIG_JAR_FROM_ABOVE io.mediachain.transactor.JournalServer path/to/config.conf`
+### RPC facade configuration
+Required properties:
+```
+io.mediachain.transactor.rpc.port: 10001
+io.mediachain.transactor.rpc.control: /path/to/rpc/killfile-dir
```
parkan, line 86: Running the facade with the invocation above and with both of these values specified still gives:

```
Exception in thread "main" java.lang.RuntimeException: Missing configuration property: io.mediachain.transactor.dynamo.baseTable
```

vyzo, line 86: The dynamo dependency was added by the streaming branch, but the README was not updated.

parkan, line 86: Should in theory still be optional for the facade though, right?

yusefnapora, line 86: I think it will revert to being unnecessary once vyzo's refactor is finished.  Datastore access was needed for the blockchain catchup, but we should be able to do without it once we stop doing catchup in the rpc service

parkan, line 86: ahh ok I see, that part was implemented while I was kind of AFK

PR #138, opened 2016-06-17
Remove l_space

parkan:
It's looking increasingly like we aren't going to use the l_space codebase soon, so let's remove it. Can always fish it out of history if necessary.


vyzo:
go for it!

PR #139, opened 2016-06-20
RpcService fixes

vyzo:
- fix dynamo config from hardcoded local dynamo
- make transactor service executor dynamically threaded
- make a dedicated execution context for serving rpc futures

### DIFF
--- a/transactor/src/main/scala/io/mediachain/transactor/RpcService.scala
+++ b/transactor/src/main/scala/io/mediachain/transactor/RpcService.scala
@@ -30,13 +30,11 @@ object RpcService {
     logger.info(s"Connecting to cluster at $cluster...")
     client.connect(cluster)
 
-    implicit val ec = ExecutionContext.global
-    // todo: dynamic
-    val executor = Executors.newFixedThreadPool(4)
-    val datastoreConfig = DynamoDatastore.Config(
-      "Mediachain",
-      Some("http://localhost:8000")
-    )
+    val threads = Math.max(4, Runtime.getRuntime.availableProcessors)
+    implicit val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(threads))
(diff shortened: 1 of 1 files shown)


bigs:
lgtm

transactor/src/main/scala/io/mediachain/transactor/RpcService.scala
```diff
@@ -30,13 +30,11 @@ object RpcService {
...
 
-    implicit val ec = ExecutionContext.global
-    // todo: dynamic
-    val executor = Executors.newFixedThreadPool(4)
-    val datastoreConfig = DynamoDatastore.Config(
-      "Mediachain",
-      Some("http://localhost:8000")
-    )
+    val threads = Math.max(4, Runtime.getRuntime.availableProcessors)
+    implicit val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(threads))
```
parkan, line 34: what ends up using this?

yusefnapora, line 34: I think this gets used by the rpc service methods.  They return `Future[SomeResponse]`, and I think they get executed on this context

vyzo, line 34: The execution context is for the futures and rpc service main body.
The extra executor is used by the streaming parts of the service.
I think we might be able to use the same executor for both, but I am not clear how that code works so I took the safe approach and kept them separate.

parkan, line 34: ah yeah `TransactorService` is the one

parkan, line 34: actually `TransactorService` takes both `executor: ExecutorService` and `implicit val executionContext: ExecutionContext`, which seems unnecessary. Can leave it be for now but something to clean up in the future.

PR #140, opened 2016-06-20
Cleanup a few l-space stragglers

yusefnapora:
removes the orientdb_migrations project dependency from Build.scala and a stray translation_engine file


vyzo:
:+1:

PR #142, opened 2016-06-21
Transactor RPC streaming rework

parkan:
Implements the transactor streaming service in the logical concurrent way:
- The backfilling behavior has been removed.
- Streaming begins at current block boundaries and absorbs transient copycat errors.
- Cleans up the threading situation

### DIFF
--- a/transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
@@ -1,10 +1,8 @@
 package io.mediachain.copycat
 
-import java.util.concurrent.ExecutorService
-import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.Executors
 
 import cats.data.Xor
-import dogs.Streaming
 import io.grpc.stub.StreamObserver
 import io.grpc.{ServerBuilder, Status, StatusRuntimeException}
 import io.mediachain.copycat.Client.{ClientState, ClientStateListener}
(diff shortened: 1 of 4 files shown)


transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -15,204 +14,313 @@ import io.mediachain.protocol.Transactor.{JournalError, JournalListener}
...
+    def extendCurrentBlock(block: JournalBlock) {
+      val newentries = block.entries.drop(state.block.length)
+      state.index = block.index
+      state.block ++= newentries
+      newentries.foreach { entry =>
+        emitEvent(TransactorService.journalEntryToEvent(entry))
+      }
+    }
+    
+    def fetchBlocks(hd: Option[Reference]) = {
```
parkan: more descriptive name than `hd` please

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -15,204 +14,313 @@ import io.mediachain.protocol.Transactor.{JournalError, JournalListener}
...
+            
+            // be tolerant of dynamo datastore transient failures
+          case Failure(err: AmazonClientException) =>
+            logger.error("AWS error", err)
+            val backoff = Client.randomBackoff(retry)
+            logger.info(s"Retrying ${ref} in ${backoff} ms")
+            Thread.sleep(backoff)
+            loop(retry + 1)
+            
+          case Failure(err: Throwable) =>
```
parkan, line 254: Don't catch `Throwable`, this will make you catch things like OOM and leave the jvm in an undefined state as your program tries chugging along. Catch `Exception` instead.

vyzo, line 254: it's just being rethrown; it's already caught by the Try construct.

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo: will fix; I'll call it chainHead

PR #147, opened 2016-06-21
Ansible GRPC playbook + other ansible tweaks

yusefnapora:
This adds a grpc playbook to checkout and build grpc from source.  

You can override the default repo by setting the `grpc_repo` variable, and the tag / commit to checkout with `grpc_checkout`.  Technically you could put a branch name for `grpc_checkout`, but that will mess with the idempotency check.  When the playbook installs grpc it drops a file at `/opt/grpc/.grpc_version_installed{{ grpc_checkout }}` to avoid rebuilding after it's already installed.  So if you use a branch name, it won't trigger a rebuild if the branch tip changes.

Also does a few other things:
- adds the `-sharedDb` flag to the local dynamo db service, and fixes the idempotency check on downloading the dynamo jar
- moves the `apt upgrade` task from the transactor playbook to `transactor-provisioning.yml` b/c it was messing up my vmware box when the kernel got updated
- adds some missing dependencies to `transactor-client-provisioning.yml` (libjpeg-dev and libpng-dev)
- adds the new grpc role to `dev.yml` and `transactor-client-provisioning.yml`, pinned to the latest commit on master -- [be223358795b645416e801727f4d3fcad3d73964](https://github.com/grpc/grpc/commit/be223358795b645416e801727f4d3fcad3d73964)

It would be nice if the grpc role could optionally install to a virtualenv, but I didn't want to get carried away :)

Also, the grpc playbook does pretty much everything as root, since it's set to checkout the code to `/opt/grpc` by default and needs write permissions.  I'd like to just check it out to the home dir instead, but the username varies between vagrant and the deployment env, and I don't know how to abstract that in ansible.  I guess I could add a `current_user` variable to the playbook, but it seems like that's probably built in to ansible and I just haven't found it.

### DIFF
--- a/provisioning/playbooks/roles/grpc/tasks/main.yml
+++ b/provisioning/playbooks/roles/grpc/tasks/main.yml
@@ -1,0 +1,40 @@
+
+- name: install grpc base dependencies
+  apt: name={{item}}
+  become: yes
+  become_user: root
+  with_items:
+      - python-dev
+      - python-pip
+      - libtool
+      - autoconf
+
+- name: clone grpc git repo
+  git: repo=https://github.com/grpc/grpc.git
+       dest={{ grpc_source_dir }}
+       version={{ grpc_version }}
+  become: yes
+  become_user: root
+
+- name: install gprc python dependencies
+  command: pip install -r requirements.txt
+  args:
+    chdir: "{{ grpc_source_dir }}"
+  become: yes
+  become_user: root
+
+- name: install grpc python module
+  command: env GRPC_PYTHON_BUILD_WITH_CYTHON=1 pip install .
+  args:
+    chdir: "{{ grpc_source_dir }}"
+    creates: "/usr/local/lib/python2.7/dist-packages/grpcio-{{ grpc_version_number }}.egg-info"
+  become: yes
+  become_user: root
+
+- name: build and install grpc
+  shell: make install && rm -f ./.grpc_version_installed_* && touch ".grpc_version_installed_{{grpc_version}}"
+  args:
+    chdir: "{{ grpc_source_dir }}"
+    creates: "{{grpc_source_dir}}/.grpc_version_installed_{{grpc_version}}"
+  become: yes
+  become_user: root
--- a/provisioning/playbooks/roles/grpc/tasks/main.yml
+++ b/provisioning/playbooks/roles/grpc/tasks/main.yml
@@ -9,32 +9,40 @@
       - libtool
       - autoconf
 
-- name: clone grpc git repo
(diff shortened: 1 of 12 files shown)


provisioning/playbooks/transactor-provisioning.yml
```diff
@@ -4,6 +4,11 @@
     - transactor
 
   tasks:
+  - name: update system
+    apt: upgrade=yes
+    become: yes
```
vyzo: becoming is unnecessary since the whole playbook needs to be run with -b

provisioning/playbooks/roles/transactor/tasks/main.yml
```diff
@@ -1,7 +1,3 @@
-- name: update system
```
vyzo, line 1: I think we should probably have this in a role? We need it in all the proviosioning scripts.

provisioning/playbooks/transactor-client-provisioning.yml
```diff
@@ -2,11 +2,18 @@
 - hosts: transactor-client-provisioning
   roles:
     - transactor
+    - {role: grpc, grpc_checkout: "be223358795b645416e801727f4d3fcad3d73964" }
```
vyzo, line 8: need system upgrade here now that it has been moved out of transactor!

vyzo:
Can you move the system upgrade in a role that is used by both transactor-provisioning and transactor-client-priviosioning? As it stands right now the client-provisioning playbook is missing the system update!

provisioning/playbooks/roles/transactor/tasks/main.yml
yusefnapora, line 1: good call

provisioning/playbooks/roles/grpc/tasks/main.yml
```diff
@@ -0,0 +1,48 @@
...
+  apt: name={{item}}
+  become: yes
+  become_user: root
+  with_items:
+      - python-dev
+      - python-pip
+      - libtool
+      - autoconf
+
+- name: update pip
```
vyzo: you need to install it first!
Why not make the pip install + upgrade a separate role and reuse that (also in the client provisioning script)

yusefnapora: pip gets installed in that apt step above.  But I agree about splitting it out into a role.  Maybe along with python-dev and python-virtualenv?

vyzo: yeah, that might be a good idea as well.

vyzo:
:+1:

PR #151, opened 2016-06-23
Modify the Transactor lookup API to distinguish nil chains from bad references

vyzo:
The lookup API currently does not allow a client to distinguish a nil chain from a canonical reference unknown to the journal.
This PR modifies the lookup API to distinguish the two cases by returning an Xor and mapping the option to the protobuf interface.


yusefnapora:
👍

PR #155, opened 2016-06-27
Unified transactor service script and configuration, testnet tooling

vyzo:
- Creates a new provisioning role transactor_server, used by transactor-provisioning.
  The role provides a unified transactor service interface through the `transactor-svc` script and generates the necessary configuration files from templates.
- Adds testnet tooling: playbooks to start/kill+wipe testnets
- Refactoring: old transactor role has been renamed to transactor_deps as this is exactly what it was providing


yusefnapora:
Looking good 👍   I really like having the templates for the server config files.

provisioning/playbooks/roles/transactor_server/templates/DatastoreRpcService.conf.j2
```diff
@@ -0,0 +1,3 @@
+io.mediachain.datastore.rpc.port: 10002
```
parkan, line 1: seems a little strange to hardcode these

vyzo, line 1: I think it's the right place to hardcoded, as we need to specify them somehow.
We could add a variable, but I don't think we need it rn.

Issue #156, opened 2016-06-27
Protocol versioning


parkan:
TODO: versioning at libp2p level

vyzo:
I don't think that meta-schema versioning can be strict; we can't retroactively change extant records.

parkan:
edited to remove no longer relevant layers

PR #157, opened 2016-06-27
Put protobuf in subtree

parkan:
Partially addresses #156

Subtree is better than submodule, because you get to mostly work with it as normal files, except when transmitting or receiving remote changes (which requires `git subtree push -P protocol/src/main/protobuf` instead of just `git push`)

More tooling to come


bigs:
LGTM :+1:

PR #159, opened 2016-06-28
transactor: Show log timestamps by default

vyzo:
Adds the magic resource file for the slf4j logger we are using.


yusefnapora:
👍

PR #160, opened 2016-06-29
transactor-client-provisioning fixes

vyzo:
- playbook needs to install packages, so it needs to become on its own without a -b argument
- grpc role was hosed, as a Makefile is necessary for the protobuf build and there isn't any in the checked out sources; it needs to be generated by running the added magic incantations


parkan:
Hopefully won't need much of this soon, but merging for now

PR #161, opened 2016-06-30
TransactorService Streaming: Fix a race condition with better state synchronization

vyzo:
While reviewing the code I noticed that there is a window between currentBlock and its completion which could lead to some JournalEvents to be duplicated and corrupt state, even though they were superseded by the current block.

This patch fixes the race condition by tracking the next entry index and ignoring blocks events with smaller indices.
It turns out that this requires tracking the index in JournalBlockEvents, as emitted by the copycat StateMachine.
The change is internal to the transactor components and not propagated in the TransactorService interface.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
@@ -26,7 +26,6 @@ class TransactorListenerState {
   var block: Buffer[JournalEntry] = new ArrayBuffer
   var blockchain: Option[Reference] = None
   var streaming: Boolean = false
-  var recovery: Set[Reference] = Set()
   
   def isEmpty = (index == -1)
 }
@@ -102,10 +101,10 @@ extends ClientStateListener with JournalListener {
       }})
   }
   
-  override def onJournalBlock(ref: Reference) {
+  override def onJournalBlock(ref: Reference, index: BigInt) {
     exec.submit(new Runnable {
       def run {
-        withErrorLog(journalBlockEvent(ref))
+        withErrorLog(journalBlockEvent(ref, index))
       }})
   }
   
@@ -140,10 +139,14 @@ extends ClientStateListener with JournalListener {
     }
   }
   
-  private def journalBlockEvent(ref: Reference) {
-    if (state.streaming && !state.recovery.contains(ref)) {
+  private def journalBlockEvent(ref: Reference, index: BigInt) {
+    if (state.streaming && index >= state.index) {
+      if (index != state.index) {
+        logger.warn(s"Block skipped entries ${state.index} -> ${index}")
+      }
       state.blockchain = Some(ref)
       state.block.clear()
+      state.index = index
       emitEvent(TransactorService.journalBlockReferenceToEvent(ref))
     }
   }
@@ -222,7 +225,6 @@ extends ClientStateListener with JournalListener {
       def loop(ref: Reference, blocks: List[JournalBlock])
       : List[JournalBlock]
       = {
-        state.recovery += ref
         val block = getBlock(ref)
         if (block.chain == state.blockchain) {
           block :: blocks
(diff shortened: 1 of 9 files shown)


transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -129,17 +129,24 @@ extends ClientStateListener with JournalListener {
...
   
   private def journalCommitEvent(entry: JournalEntry) {
-    if (state.streaming) {
+    if (state.streaming && entry.index >= state.index) {
+      if (entry.index != state.index) {
+        logger.warn(s"Commit skipped entries ${state.index} -> ${entry.index}")
+      }
       state.block += entry
-      state.index = entry.index
+      state.index = entry.index + 1
```
parkan, line 137: why'd this change?

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -129,17 +129,24 @@ extends ClientStateListener with JournalListener {
   }
   
   private def journalCommitEvent(entry: JournalEntry) {
-    if (state.streaming) {
+    if (state.streaming && entry.index >= state.index) {
+      if (entry.index != state.index) {
+        logger.warn(s"Commit skipped entries ${state.index} -> ${entry.index}")
```
parkan, line 134: This sounds like an open interval (inclusive), but it's actually a closed interval, right?

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
```diff
@@ -129,17 +129,24 @@ extends ClientStateListener with JournalListener {
...
-      state.index = entry.index
+      state.index = entry.index + 1
       emitEvent(TransactorService.journalEntryToEvent(entry)) 
     }
   }
   
-  private def journalBlockEvent(ref: Reference) {
-    if (state.streaming) {
+  private def journalBlockEvent(ref: Reference, index: BigInt) {
+    if (state.streaming && index >= state.index) {
```
parkan, line 143: so the index is checked in 2 different places, but we're guaranteed no race between these, right? can these two events happen concurrently?

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo, line 137: to be consistent with the block index, which is the next seqno expected.

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo, line 134: It really should be == to the state.index, but I am programming it defensively.

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo, line 143: No, they are serialized in the exec executor.

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
parkan, line 134: I mean the error message, it suggests that entries at `state.index` and `index` were skipped, while they actually were not, only the entries between them were

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
parkan, line 137: gonna trust your math on this one, don't have a clear picture of event sequencing here

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
parkan, line 143: Great :+1:

transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
vyzo, line 134: The error message indicates that our index jumped, so there must be some skipped entries.
(again, this should not happen but famous last words)

PR #162, opened 2016-06-30
cbor tag support

yusefnapora:
This replaces the `CborAST` type `CTag` with `CTaggedValue`, which wraps another `CValue`.  It turns out that cbor-java doesn't emit `Tag` items directly when decoding, so `CTag` wasn't needed.  Instead, each cbor-java `DataItem` has a `.tag` member which is `null` if there's no tag.

Anyway, this seems to correctly handle tagged values, and I updated the cbor serialization for `MultihashReference`s to emit the multihash as `CBytes` tagged with the IPLD link tag value (258), instead of the `CMap` with `@link` field we were using.

The deserialization code can still handle the old `@link` maps, so references encoded with the old serializers will still work.

This will need client support, since the current python client is expecting refs to be `@link` maps.
I'd also like to test this with the js-ipld implementation to see if these links actually resolve.


protocol/src/main/scala/io/mediachain/protocol/CborSerialization.scala
```diff
@@ -693,8 +693,8 @@ object CborSerialization {
 
     override def fromCValue(cValue: CValue): Xor[DeserializationError, MultihashReference] = {
       cValue match {
-        case CTaggedValue(tag, CBytes(value)) if tag == MultihashReference.cborLinkTag =>
-          MultiHash.fromBytes(value)
+        case CTaggedValue(tag, CString(value)) if tag == MultihashReference.cborLinkTag =>
+          MultiHash.fromBase58(value)
```
parkan: hmm I wonder if this is a js-ipld thing

yusefnapora: Yeah, I can't tell whether this is a design decision or just an implementation bug 😄   The current js ipld implementation assumes that byte strings are multiaddrs, so if you give it a multihash instead it will fail to decode.  But it parses base58 multihash strings just fine and resolves them as ipfs links.

vyzo: Yeah, this thing can have quite detrimental effect in object sizes.

yusefnapora: hmm, just read through [the spec](https://github.com/ipfs/specs/tree/master/ipld#serialised-cbor-with-tags), and it seems the js behavior is correct:
- If the link is a valid multiaddress and converting that link text to the multiaddress binary string and back to text is guaranteed to result to the exact same text, the link is converted to a binary multiaddress stored in CBOR as a byte string (major type 2).
- Else, the link is stored as text (major type 3)

So we might need to use multiaddr after all if we want efficient storage.  The multiaddr header only adds about 3 bytes.  I've got the scala multiaddr port mostly working, but i need to do an error-handling pass and run some tests.

parkan: Ok that makes sense :)

I'm kinda inclined to just make a multiaddress branch w/o dynamo support. We could also potentially try to specify dynamo addresses in another way, like using the REST API?

yusefnapora:
so, i got the [multiaddr branch](https://github.com/mediachain/scala-multihash/tree/yn-multiaddr) of the multihash project working a little bit ago.  I think if we want to use it it should probably be split into its own project though. 

But it occurred to me that all of our refs will have the exact same three byte header (ipfs code + size), since we always use SHA2-256 to encode our hashes, so the size byte will always be the same.  So we could have multiaddr compatible links by just prefixing our raw multihashes with those three bytes, and checking for their presence when we decode.  That would be enough to have compact IPLD-compatible links, and it would be dead simple to implement.

vyzo:
that sounds reasonable.

parkan:
@yusefnapora sounds great, I was wondering why we needed the full machinery. So string `ipld://Qm...` encodes to which leading byte? From [here](https://github.com/jbenet/js-multiaddr/blob/369d996576c3d15e357b8187dea32f5cbe08cc04/src/protocols.csv) it kinda seems like it should be 421, is that right?

yusefnapora:
it is 421, but encoded as a protobuf-style Varint.

yusefnapora:
okay, last commit adds the fixed multiaddr header.  this still needs client support before we can merge it in though.  @parkan do we want to get this in before the testnet launch?

parkan:
I'm kinda inclined to hold off on that for a moment, we can put it into 0.2.0 release

parkan:
Unless you think there's a concrete benefit of getting this in as early as possible?

protocol/src/main/scala/io/mediachain/protocol/Datastore.scala
```diff
@@ -42,10 +42,22 @@ object Datastore {
...
+    // IPLD uses cbor tags to indicate a traversable link
+    // if the tagged value is a byte array, it will be interpreted
+    // as a multiaddr (see https://github.com/jbenet/multiaddr)
+    val cborLinkTag: Long = 258
+
+    // prefix multihash bytes with a multiaddr header
+    // the header value is 0xA50322 which is the ipfs multiaddr code 0xA503
+    // (421, encoded as a protobuf-style varint)
+    // + the length of a sha2-256 multihash 0x22 (34 bytes)
+    val multiaddrHeader: Array[Byte] = Array[Byte](-91, 3, 34)
```
vyzo: can we have the bytes as hex digits somehow?

parkan: I think you can put in `0x...` hex literals yeah

autoencoder: Yeah if we switched to hex digits for all APIs too it'd be really convenient. Our JSON serializers really don't like raw byte strings.

yusefnapora: The scala hex literals are interpreted as `Int`, which is kind of annoying, since e.g. `0xA5` is interpreted as outside the (signed) range of a `Byte`.  So you have to manually cast it to `Byte`:  

```
scala> 0xa5 
res0: Int = 165
scala> val b: Byte = 0xA5 
Main.scala:45: type mismatch;
 found   : Int(165)
 required: Byte
val b: Byte = 0xA5
              ^
Compilation Failed
scala> val b: Byte = 0xA5.asInstanceOf[Byte] 
b: Byte = -91
```

vyzo: this is just sad.

parkan: I'm not mad I'm just disappointed

parkan:
Ditto here, let's review tmw

PR #163, opened 2016-07-01
Transactor: assorted fixes

vyzo:
Several fixes given the ingestion bug experience.

Server related issues:
- make server shutdown robust to exceptions
- make transactor-svc script forcibly kill non-cooperating daemons

Copycat issues:
- bump copycat version (fixes bugs)
- decouple the state change event dispatch from the copycat thread in Client
- make operations eventually fail in periods of cluster unavailability in Client
- adjust cluster timings to be less aggressive: 1s for heartbeat (instead of 250ms), 3s for election (instead of 500ms)


transactor/src/main/scala/io/mediachain/copycat/Client.scala
```diff
@@ -117,15 +122,27 @@ class Client(client: CopycatClient) extends JournalClient {
     }
   }
   
+  private def emitStateChange(stateChange: ClientState) {
```
parkan, line 125: :+1:

parkan:
Looks good!

PR #165, opened 2016-07-05
wip - publish to sonatype

yusefnapora:
So far, just resolves the scala-multihash dependency from sonatype instead of using a git project dependency.  I figured that'd be a better project for "learning the ropes" for this crazy publication workflow.

Next up:
- [x] Add publication metadata to base settings
- [x] Publish protocol and transactor projects to sonatype
- [ ] figure out how to automate deployment in travis CI

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -160,36 +160,34 @@ object MediachainBuild extends Build {
       // to indicate a build that's "in-between" releases
       git.uncommittedSignifier := None,
 
-      // If there are no tags prior to the head commit, set the version
-      // number to 0.1.0-${sha}-SNAPSHOT where ${sha} is the first 8 chars
-      // of the head sha-1 hash.
+      // If there are no matching tags prior to the head commit, set the version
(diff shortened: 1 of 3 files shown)


parkan:
Need to update https://github.com/mediachain/mediachain/blob/ebfd6328ba2b5202d9a4a83d3709e7c48204f6fc/provisioning/playbooks/roles/transactor_server/files/transactor-svc to remove version, what else?

yusefnapora:
Oh, I updated that in the last commit: https://github.com/mediachain/mediachain/commit/6fce5e0222057ca0e8bf95055ac691ed75b8cf86

Now it always generates `transactor-assembly.jar` and the provisioning script uses it without the version number

parkan:
Oh oops didn't see the second file in that commit, are we good to merge this now then?

yusefnapora:
I think so, yeah.  It shouldn't break anything :)  We'll circle around to automated snapshot deployment when we figure out the e2e testing next week

PR #166, opened 2016-07-07
Moar transactor stability fixes

vyzo:
- Go Medieval with copycat clients; https://github.com/atomix/copycat/issues/241
- Configure threadiness for persistent datastore in transactor provisioning
- Add ntp to transactor provisioning

### DIFF
--- a/transactor/src/main/scala/io/mediachain/copycat/Client.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/Client.scala
@@ -152,15 +152,18 @@ class Client(sslConfig: Option[Transport.SSLConfig]) extends JournalClient {
   }
   
   private def reconnect() {
+    def newClient() {
+      // Copycat client state is already closed if we are reconnecting, 
+      // but also close it to shutdown its internal context
+      val klient = client
+      client = newCopycatClient()
+      withErrorLog(klient.close())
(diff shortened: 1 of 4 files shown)


transactor/src/main/scala/io/mediachain/copycat/Client.scala
```diff
@@ -169,12 +177,16 @@ class Client(sslConfig: Option[Transport.SSLConfig]) extends JournalClient {
...
-            case Failure(e) =>
+              
+            case Failure(e: Exception) =>
               logger.error("Connection error", e)
               val sleep = Client.randomBackoff(retry)
               logger.info("Backing off reconnect for " + sleep + " ms")
               Thread.sleep(sleep)
               loop(retry + 1) 
+              
+            case Failure(e: Throwable) =>
```
parkan, line 188: OOM was hitting the above branch and was leaving the jvm hosed, right?

parkan:
Well, sure is ugly but if it works it works

transactor/src/main/scala/io/mediachain/copycat/Client.scala
vyzo, line 188: no, it was happening in the journal server, not here.

PR #170, opened 2016-07-12
fix spurious dependency updates due to git versioning

yusefnapora:
changes Build.scala to only include git sha in version number if a special property is set
- if the `io.mediachain.build.versionIncludesGitSha` property is set to any value,
  the version number will include the short sha when not on a tagged
  commit.


vyzo:
:+1:

PR #171, opened 2016-07-12
TransactorService: publish error metric on insert duplicate errors

vyzo:
Add the missing error metric for the insert JournalError branch.


yusefnapora:
👍

PR #172, opened 2016-07-13
Add object metrics to transactor service

vyzo:
A counter tracks record insert/update and a gauge tracking the current index (height of the blockchain).


parkan:
Looks good, gonna merge once rebuild succeeds

PR #173, opened 2016-07-13
bump grpc to 0.15.0, sbt-scalapb to 0.5.32

yusefnapora:
You'll probably need to run either `sbt clean` or `sbt protobuf:protobufGenerate` after updating these dependencies.

The old version of sbt-scalapb we were on generates code that's deprecated by the new grpc.  The new version is compatible, but I guess since the original protos were unmodified they weren't triggering a recompilation, so it was trying to build with the old generated code.  Forcing the protos to recompile fixes the error.


vyzo:
:+1:

PR #176, opened 2016-07-14
Client provisioning updates

yusefnapora:
Makes a few changes to the client provisioning playbook:
- removes the grpc dependency, since we don't need to install from source anymore (yay!)
- adds a rocksdb playbook to install rocks into /usr 
  - i'd prefer /usr/local but I had trouble getting the pyrocksdb package to find it when installed there
- makes transactor-client-provisioning.yml depend on rocks playbook, install pyrocksdb to system pip
- adds ipfs role from ansible galaxy to transactor-client-provisioning.yml

The last one means that you'll need to run `ansible-galaxy install insanity54.ipfs` on your local machine after this is merged.

Also, one thing to note is that the ipfs daemon runs as the `ipfs` user (created by the ansible role), so if you shell in and try to run e.g. `ipfs cat some-hash` it will complain at you.  Everything seems to work via the HTTP api though, which is what we're using in the client.  You can `su` to the `ipfs` user if you need to run ipfs commands from the cli.

@vyzo want to try this out tomorrow and pull it in?


yusefnapora:
Also, vyzo, if you find that ipfs is slowing down ingestion or eating up too much cpu tomorrow, you can disable it during ingestion with the `--disable-ipfs` cli flag.  It was switched to "on by default" in https://github.com/mediachain/mediachain-client/pull/65

vyzo:
Uhm /usr installation is rather rude -- can we fix that one please?

yusefnapora:
ugh, there's a problem with the ipfs role; apparently since ansible 2.1 it's much harder to become an unprivileged user, so it fails when running `ipfs init` as the `ipfs` user.  I thought I had fixed that yesterday, but apparently not.

yusefnapora:
so the ipfs thing is an issue with ipfs, not ansible: https://github.com/ipfs/go-ipfs/issues/2748

`ipfs init` fails if stdin is not a tty 😞  it's fixed in master but not released yet.
I think I'll just remove the ipfs role for now, and we can revisit later.

parkan:
That sucks. Can you try `script --return -c "ipfs init" /dev/null`?

yusefnapora:
nice, that seems to work 👍  I'll push an update in a sec.

PR #178, opened 2016-07-18
intercept rpc calls and log unique client addresses

yusefnapora:
so far just prints the client's `SocketAddress` to the console.  needs to be wired up to the influxdb metrics logger

### DIFF
--- a/transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
+++ b/transactor/src/main/scala/io/mediachain/copycat/TransactorService.scala
@@ -1,6 +1,6 @@
 package io.mediachain.copycat
 
-import java.net.SocketAddress
+import java.net.{InetAddress, InetSocketAddress, SocketAddress}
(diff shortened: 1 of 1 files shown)


vyzo:
can you log it instead of printing so that we can actually see them?

yusefnapora:
sure, one sec :)

yusefnapora:
@vyzo, does this log output look alright:

```
[18/07/2016:15:50:32 +0000] [grpc-default-executor-0] INFO UniqueClientIP - /0:0:0:0:0:0:0:1
```

It looks like the localhost transport  at least uses ip6 - I imagine we'll see some ip4 from real-world clients also.

I'm just storing the hashCode of the addresses in a mutable `HashSet` to track unique connections.  It will lose the history if the service restarts, but that's probably fine; it doesn't seem worth the bother of persisting the set of connections.

Should I try wiring this up to influxdb also?

vyzo:
:+1:

PR #180, opened 2016-07-18
add self-hosting docs, link to from testnet docs.

yusefnapora:
Adds a `selfhosting.md` doc with instructions on building and running the transactor services.


parkan:
Looks good; I think this is about all the attention we should give to this subject rn

PR #181, opened 2016-07-19
Refactor, add unique IP logging to the datastore service

vyzo:
- refactors the IP logging interceptor out of the TransactorService
- adds IP logging to DatastoreService


yusefnapora:
👍 good call

PR #183, opened 2016-07-20
add indexer invocations

yusefnapora:
small doc update with instructions on how to start the indexer.
@parkan, should I move this to the self-hosting doc, and just put instructions for querying the deployed testnet indexer in this doc?


parkan:
Yeah I think the main doc should not concern running services yourself, looks good otherwise

PR #187, opened 2016-07-29
WIP: kittengrad

vyzo:
The Kittengrad Transactor Network Draft to sketch out next phase implementation.

### DIFF
--- a/docs/kittengrad.md
+++ b/docs/kittengrad.md
@@ -61,7 +61,7 @@ of write operations. Previously, the operations returned a commit receipt in the
 form of a `JournalEntry` immediately.
 Now they return a transaction id, which will be included in some `JournalEntry` in
 a future block streamed through the `journalStream` interface.
-The stream interface itself changes to only emit commited block events instead
+The stream interface itself changes to only emit committed block events instead
 of individual entries.
 
 ## Transactor Implementation
@@ -80,12 +80,12 @@ a p2p gossip protocol.
 ### Transactor State
 
 Each transactor maintains a local copy of the blockchain on disk, and a pointer
-to the last commited block. The transactor also maintains on disk the mediachain
+to the last committed block. The transactor also maintains on disk the mediachain
 index which corresponds to the blockchain, mapping canonical to chain references.
 The index is used for creating update transactions and providing read access
 to the mediachain.
 
-In addition to transactions commited in the blockchain, each transactor
+In addition to transactions committed in the blockchain, each transactor
 maintains a worklog of pending transactions. The blockchain index together with
 the worklog compute an aggregate index, the working index.
 
@@ -106,7 +106,7 @@ transaction to the network, and returns the transaction id to the client.
 
 If the transaction is the first one in the worklog, the transactor sets
 a timer that ensures a block confirming the transaction will eventually
-be commited with the Block Commit Protocol.
+be committed with the Block Commit Protocol.
 If the worklog size exceeds the maximum block size, and there is no
 commit in progress, the transactor initiates the Block Commit Protocol
 in order to commit some transactions off its worklog.
@@ -155,12 +155,12 @@ The BCP0 protocol operates in rounds, with each round having three phases:
 
 1. A Proposal phase, where one or more new blocks are proposed, validated and propagated by transactors
 2. A voting phase, where transactors vote for the next block
-3. A commit phase, where a block is commited.
+3. A commit phase, where a block is committed.
 
 #### Block Proposal
 
 During the block proposal phase, one or more transactors with stake in the
-form of ownership of pending transactios,  generate a block and brodcast a
+form of ownership of pending transactions,  generate a block and brodcast a
 signed block proposal together with the block data.
 
 Upon receiving a block proposal, a transactor that has not
@@ -205,7 +205,7 @@ the proposal phase.
 
 The protocol ends when enough commit messages have been received.
 If there is no majority commit at the end, the protocol again re-establishes
-network membership are starts another round. Partial failures that don't
+network membership and starts another round. Partial failures that don't
 affect block commit are also handled by network reconfiguration, but
 this time in the background.
 
@@ -221,20 +221,20 @@ be reconciled with the blockchain merging approaches discussed later.
 When the system is operating at steady state, it is constantly under
 some write load.  While the system is commiting a block with the block
 commit protocol, it is also accumulating newer transactions that will
-need to be commited in a subsequent block.
+need to be committed in a subsequent block.
 
-Once a block has been commited, transactors examine their resulting
+Once a block has been committed, transactors examine their resulting
 worklog, which is the queue built up while the block was being
-commited.  Depending on the load, this may be less a full block, in
+committed.  Depending on the load, this may be less a full block, in
 which case the next block commit execution follows the timer of the
 first pending transaction.
 
 Under higher write load however, this will exceed the block size,
 necessitating an immediate re-execution of the block commit
-protocol. In order to minimze contention in block proposal, which
+protocol. In order to minimize contention in block proposal, which
 consumes bandwidth for block propagation, we want to utilize a
 randomized wait period before a transactor initiates block commit by
-proposing a new block. On the same time we want to minimize the time
+proposing a new block. At the same time we want to minimize the time
 it takes for the first proposed block, which initiates protocol
 action.  We can modulate the delay by determining the wait period with
 a distribution dependent on transactor stake. In this way, transactors
--- a/docs/kittengrad.md
+++ b/docs/kittengrad.md
@@ -159,8 +159,8 @@ The BCP0 protocol operates in rounds, with each round having three phases:
 #### Block Proposal
 
 During the block proposal phase, one or more transactors with stake in the
-form of pending transactios,  generate a block and brodcast a signed block proposal
-together with the block data.
+form of ownership of pending transactios,  generate a block and brodcast a
+signed block proposal together with the block data.
 
 Upon receiving a block proposal, a transactor that has not
 yet proposed a block can validate and endorse the block by broadcasting
@@ -180,7 +180,7 @@ for it.
 The choice should be deterministic, so that all correct transactors should
 make the same decision in the absence of network failures:
 - If there is only one block proposed, the choice is obvious.
-- If there are more, then the transactor votes for the block with the higest index.
+- If there are more, then the transactor votes for the block with the highest index (most transactions).
 - If two blocks have the same height, then the transactor chooses by comparing block hashes.
 
 The transactor then broadcasts a Vote message, indicating the chosen block and awaits
@@ -252,7 +252,7 @@ So far we have developed the protocol using strict writes: transactors
 immediately write to the datastore and propagate transactions that
 refer to objects by references. This is an optimistic approach to
 concurrency, as it assumes little contention from concurrent writes,
-which are resolved by transaction invalidation.
+which are then resolved by transaction invalidation.
 
 An alternative approach is to treat updates as writes into partially
 ordered sets. There is no total order for updates, but each transactor's
--- a/docs/kittengrad.md
+++ b/docs/kittengrad.md
@@ -87,7 +87,7 @@ to the mediachain.
 
 In addition to transactions commited in the blockchain, each transactor
 maintains a worklog of pending transactions. The blockchain index together with
-the worklog transactions compute an aggregate index, the working index.
+the worklog compute an aggregate index, the working index.
 
 Read requests are serviced by the working index, while write operations are
 initiated on the working index and completed through the transaction protocol.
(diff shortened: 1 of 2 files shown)


parkan:
city-of-cats-cross-stitch-pattern-5

docs/kittengrad.md
```diff
@@ -0,0 +1,151 @@
...
+backoff and retry with a randomized exponential backoff or similar mechanism.
+
+The transactor will see through the completion of accepted transaction at a best
+effort basis, which in the absence of transactor failure should result in the eventual
+commit of an accepted transaction.
+The client may wait to confirm a transaction through a streamed block,
+and potentially retry transactions aborted because of transactor failure.
+
+The interface changes from the v0.1 interface are related to the asynchronous nature
+of write operations. Previously, the operations returned a commit receipt in the
```
parkan, line 60: FWIW, the tx receipt for Tendermint looks like this: https://github.com/eris-ltd/eris-db/blob/develop-tmsp/rpc/core/types/responses.go#L102

Not entirely clear how you're thinking about this from the text, but I think the idea is that you get a tx id/hash back immediately upon acceptance, then listen for new blocks and see if your tx is in them as the confirmation mechanism.

vyzo, line 60: It's exactly the same thing; get the transaction id back and confirm through emitted blocks.

parkan, line 60: Cool, just wanted to get it clear 😄

docs/kittengrad.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Transaction Protocol
+#### Transaction Initiation
+
+In order to perform a write operation, the transactor checks the validity
+of the operation against the working index. Duplicate inserts are rejected,
+as are updates for unknown canonicals. It also checks its worklog backlog,
+against some safety threshold.
+
+The transactor then generates a transaction id and adds the data to the
```
parkan, line 103: Need to specify how these are generated; presumably, hash of timestamp/nonce/contents

vyzo, line 103: Yeah, that's a pretty decent mechanism. I am thinking of deterministically generating insert transaction ids so that we can merge duplicates that have been concurrently accepted. It's mentioned a little further down, perhaps I should specify the update transaction id generation there.

docs/kittengrad.md
```diff
@@ -0,0 +1,151 @@
...
+The block is stored and becomes the current blockchain pointer, while
+the blockchain index is updated in place. The worklog of remaining pending
+transactions is then revalidated against the blockchain index.
+Invalid transactions are rejected, and it is the responsibility of the originating
+transactor to replay them.
+
+In this way, the block commit provides a conflict resolution mechanism
+for concurrent updates. Inserts may use a deterministic transaction id,
+based on the canonical reference, which allows merging and confirmation
+of duplicate inserts without a replay. Updates however are dependent
```
parkan, line 138: For the general case, g-set CRDTs can handle this w/o rewrite per se (write must be originally submitted as set add). For the case where ordered self-overrides are allowed, 2P or LWW sets (w/per-client vector clocks, probably) have the desired effect.

docs/kittengrad.md
parkan, line 103: Hmm yeah gotta think about this esp if we try a CRDT solution like I suggested below

docs/kittengrad.md
vyzo, line 138: I originally had a section for a potential CRDT solution using sets, but they do have to be eventually written out during block proposal. The difference is that the entire data of the transaction has to be passed around through the transactor network instead of a datastore reference.

parkan, line 138: Hmm you don't think the set can just consist of "inner" chain object references (i.e. pointer to `meta` field and everything below)?  That would get wrapped into a full chain cell during block commit.

vyzo, line 138: I considered detaching the meta from the cons cell, but we still need to write the cell. Plus it adds round-trip times when reading the object.

vyzo, line 138: Added section discussing the CRDT approach with delayed writes in https://github.com/mediachain/mediachain/pull/187/commits/133782cba895379ccd39cd735c3644576b6b4722

docs/kittengrad.md
```diff
@@ -139,13 +139,81 @@ of duplicate inserts without a replay. Updates however are dependent
...
+
+#### Voting
+
+Once the proposal phase has finished, each node chooses a valid block and votes
+for it.
+
+The choice should be deterministic, so that all correct transactors should
+make the same decision in the absence of network failures:
+- If there is only one block proposed, the choice is obvious.
+- If there are more, then the transactor votes for the block with the higest index.
```
parkan: highest index = most transactions? presumably we have to fall back to this in the absence of difficulty

docs/kittengrad.md
```diff
@@ -139,13 +139,81 @@ of duplicate inserts without a replay. Updates however are dependent
...
+#### Block Commit
+
+Each transactor tallies the votes, and if the majority of transactors that were part
+of the round have voted for the same block, the transactor goes forward and commits
+the block to disk.
+The transactor then broadcasts a commit message and waits to hear
+from the live majority before ending the protocol once enough commit
+messages are received. 
+
+If there are not enough votes for a block, then there must be some failed nodes.
```
parkan, line 201: So each transactor must maintain a complete(ish) peer list for both the voting AND confirmation? I wonder if there is a way to do this with less state

docs/kittengrad.md
```diff
@@ -139,13 +139,81 @@ of duplicate inserts without a replay. Updates however are dependent
...
+
+The BCP0 protocol operates in rounds, with each round having three phases:
+1. A Proposal phase, where one or more new blocks are proposed, validated and propagated by transactors
+2. A voting phase, where transactors vote for the next block
+3. A commit phase, where a block is commited.
+
+#### Block Proposal
+
+During the block proposal phase, one or more transactors with stake in the
+form of pending transactios,  generate a block and brodcast a signed block proposal
```
parkan: pending transactions = transactions submitted _directly_ to that node, not rebroadcast txes, right?

parkan:
One thing that may be illuminating is comparing our protocol to existing transaction mining/validation/witnessing approaches -- I think with this piece in particular comparative analysis is very useful. It would also help us think about the steps between this BCP0 and a staked approach.

docs/kittengrad.md
vyzo: Yeah, the block retains its index in the data model, so a valid block with a higher index will include more transactions. Note that part of the validation is ensuring it has the correct blockchain.

docs/kittengrad.md
vyzo, line 201: Yeah, membership is known by all nodes in BCP0.
We can potentially simplify the commit phase to remove the wait, but it does serve a purpose in latent error discovery

docs/kittengrad.md
vyzo: Yes, originating transactions. Not sure how to word it better.

docs/kittengrad.md
vyzo: Added a little note that the higher index implies more transactions.

docs/kittengrad.md
vyzo: Restated as "in the form of ownership of pending transactions" to be clearer.

docs/kittengrad.md
vyzo, line 201: Added a paragraph about commit confirmation ellision in https://github.com/mediachain/mediachain/pull/187/commits/e187cb7dbd34a828856d42c3033eac3a1b8f5e8c

autoencoder:
High-level questions for a new blockchain design:
- Which resource(s), if you have enough of it, allows you to control the network with high probability. Or, which network(s), if infiltrated allows you to control the greater network?
- Are there already nefarious organizations that control vast amounts of the chosen resource? E.g. IP addresses, starting a new blockchain that's compatible with existing Bitcoin mining equipment.
- How does it ensure penalties for bad behaviors? (May be a long list of bad behaviors and penalty mechanisms.)
- What are typical sequences of API calls which applications built on top of this will use? E.g. API calls for two parties doing a Bitcoin transaction.
- What are typical sequences of API calls, for types of attacks that this blockchain aims to mitigate? E.g. a Bitcoin double spend attempt between two parties.
- What particular scaling tradeoffs / decisions are we aiming for from day-1? (Maximizing bulk throughput speed, supporting huge volume of artifacts, cost of making huge number of artifacts available on the network, cost of actually distributing a huge number of artifacts through out the network, dozens of other design tradeoffs.) Think CAP, but extended to dozens of choices. Unlikely IMO that big changes to these decisions can be effectively patched in later.
- Advanced - what social or economic factors make the blockchain grow stronger (instead of more vulnerable) as it grows and becomes more valuable?
- Advanced - As the relative cost and availability of resources such as network bandwidth, cpu speed, storage speed, etc. change over the years, how do these relative changes between resources affect the blockchain in the long term?

parkan:
Yeah, I think these are (good!) mostly "phase 3" questions except for

> What are typical sequences of API calls which applications built on top of this will use? E.g. API calls for two parties doing a Bitcoin transaction.

and

> What particular scaling tradeoffs / decisions are we aiming for from day-1? (Maximizing bulk throughput speed, supporting huge volume of artifacts, cost of making huge number of artifacts available on the network, cost of actually distributing a huge number of artifacts through out the network, dozens of other design tradeoffs.) Think CAP, but extended to dozens of choices. Unlikely IMO that big changes to these decisions can be effectively patched in later.

vyzo:
Yeah, agreed.

> What are typical sequences of API calls which applications built on top of this will use? E.g. API calls for two parties doing a Bitcoin transaction.

Well, we need more experience with the system in order to observe actual usage patterns. Typically, when we ingest a new artefact we have two inserts (entity + artefact) followed by an update for the creation cell. After ingestion, we can expect isolate updates on some chain.

> What particular scaling tradeoffs / decisions are we aiming for from day-1

I think we want to be maximizing throughput initially, hence the focus on block commit chaining.

docs/kittengrad.md
```diff
@@ -248,5 +248,29 @@ commit protocol.
 
 ### Delayed Writes
 
-TBD
+So far we have developed the protocol using strict writes: transactors
```
parkan, line 258: :+1:

parkan:
One more, somewhat higher level question: it seems like we will definitely need a few network level features:
- NAT traversal
- Peer discovery
- Multicast/peer routing
- Health checks
- Join notificaitons

Netty doesn't really provide any of these out of the box (other than multicast, potentially) so we would have to use 3rd party libs or roll our own. It's worth considering that libp2p provides all of these capabilities and is under rapid development; it's also kind of nice for the consensus layer and the datastore (IPFS) to have similar and compatible capabilities.

vyzo:
We don't have to use netty or even scala for the implementation; I have been considering libp2p as well.

Until we open up the transactor membership for external transactors we don't have to deal with NAT traversal.  When we open up we'll be in the peculiar situation of desiring internal routing between transactors in the private network, which is potentially tricky to solve.

parkan:
Yeah, we will want to form an overlay network with internal routing. I think the libp2p model will support this.

AFDudley:
> The Kittengrad architecture replaces the Copycat cluster with a peer-to-peer network of transactors, who independently accept and propagate transactions and synchronize with the generation of new blocks.

I am extremely skeptical this will work, because:
Transactors must be canonical over some set, to avoid collision based DoS (initiated by either clients or transactors).
Thus, transactors should be FT or BFT clusters themselves.

AFDudley:
updateChain may require handoff protocol if clients are talking to non-canonical transactors.

Each CanonicalRef can only modified (via updateChain) by a single transactor within a given block. This needs to be enforced at the application/client level to avoid merging blocks. This should also be enforced in the BCP0 layer, to stop malicious transactors.

This is how the interface should be modified to allow for CanonicalRef binding to a given transactor:
insertCanonical(data) : (TxID, CanonicalREF, [ImpliedTransactorID])
updateChain(CanonicalRef, [ImpliedTransactorID], data) : TxID
handoffCanonical(CanonicalRef, [ImpliedTransactorID], newTransactorID) : TxID

At the BCP layer, only if the Tx describing the handoff is successful will later attempts to update a given CanonicalRef from the ‘newTransactorID’ succeed.

handoffCanonical is stored in blocks to prevent attacks from malicious transactors.

AFDudley:
Transactors putting up a bond to be canonical over a subset of CanonicalRefs might make sense here.
If you wait until later, you're going to be storing nonsense data (attacker spam) in the blocks.
〰️ 〰️ 〰️ 
This relates back to the conversation we had at MIT regarding there actually being three different networks, IIRC. Trusted/Institutional transactors, public transactors, and I've forgotten the third one, I think it might have been a decentralized storage network.

Am I correct in assuming that transactors are each managing their own CRDT, which then gets merged into an ordered BCP0 block?

I am (still) confused about the relationship between BCP0, a blockchain and CRDTs. The blocks in a blockchain provide a canonical ordered view of history and (depending on the implementation) simply looking at the last block can provide a conflict free snapshot. But it seems to me in the model that you're describing end-user-clients themselves realize a conflict free canonical ordering based on a set of user specified parameters that allow them to resolve conflicts in the RDT generated by BCP0, since it's trivial for an attacker to generate conflicting updateChain messages.
〰️ 〰️ 〰️ 
it is also possible to virtualize canonical transactors, in much the same way nodes are virtualized in Cassandra. (https://docs.datastax.com/en/cassandra/2.2/cassandra/architecture/archDataDistributeVnodesUsing.html)

vyzo:
Hey @AFDudley , a couple of quick comments as you are raising interesting issues for the long term view of the mediachain. 

> Transactors must be canonical over some set, to avoid collision based DoS (initiated by either clients or transactors).
> Thus, transactors should be FT or BFT clusters themselves. 

We are not partitioning the canonical space at this phase, although this is something we are considering in the long term. Nonetheless, the idea of having interconnected BFT clusters of transactors operated by independent stakeholders is part of our large scale world-view.

As an aside, I think you are taking the doc out of context a bit. 
The kittengrad design is more concerned with implementing the initial cluster if you like, as we intend it to be the first step towards the large scale architecture and it suffices to be simply FT (not BFT).
We want to iterate and build so that we can grow the system organically.

> updateChain may require handoff protocol if clients are talking to non-canonical transactors.

We don't have the notion of canonical transactors as we don't have partitioning at this level, so there is need for a handoff protocol (yet).

> Each CanonicalRef can only modified (via updateChain) by a single transactor within a given block
> ...

This is not true -- we allow for concurrent update transactions and we have two alternatives in order to resolve conflicts: either through invalidation (strict writes) or by delayed writes which allows merging of concurrent updates.

> I am (still) confused about the relationship between BCP0, a blockchain and CRDTs.

The blockchain provides an ordered canonical view for the state of the world up to the last commited block.
At any given moment there are pending transactions, asssuming the system is under some write load.
BCP0 is tasked with synchronizing the transactors to commit the next block by packaging pending transactions. 
The CRDT part comes in the management of the worklog of pending transactions, if we go forward with a delayed write implementation. We can treat pending transactions as a mapping of canonical to partially ordered sets of updates, which commute modulo transactor-relative order. The commute space is simply the space of possible interleavings preserving transaction partial order

AFDudley:
@vyzo Thanks for clarifying. I should have said earlier the document is good and easy to read which is deeply appreciated, thanks.

> The kittengrad design is more concerned with implementing the initial cluster if you like, as we intend it to be the first step towards the large scale architecture and it suffices to be simply FT (not BFT).

For me this was the major lesson learned from BigChainDB: It's very difficult to apply BFT to a FT system after the fact. So, I generally spec out a full BFT system and then work back to FT. This issue will appear in some of my following comments.

> This is not true -- we allow for concurrent update transactions and we have two alternatives in order to resolve conflicts: either through invalidation (strict writes) or by delayed writes which allows merging of concurrent updates.

Can you please explain how either of these strategies is viable in an adversarial environment? Invalidation can be DoS vector, depending on how transactions are priced.
Merges either require a centralized authority, or you're basically recreating normal FT DB write contention and in the process destroying the validity of the transaction receipts generated by the transactors and provided to the clients. (If I am misunderstanding and these receipts are never generated, or if that's what you mean by 'delayed write', apologies. 😄 )

> The CRDT part comes in the management of the worklog of pending transactions, if we go forward with a delayed write implementation. We can treat pending transactions as a mapping of canonical to partially ordered sets of updates, which commute modulo transactor-relative order. The commute space is simply the space of possible interleavings preserving transaction partial order

I don't understand how this make sense in the FT context as opposed to the BFT context. If it's FT we can assert the pending transactions will never be write-contentious and thus we need no logic to resolve ordering. But if it's BFT we should assume a significant percentage of transactions will be write-contentious and created at the greatest inconvenience to the transactors. (Which goes back to my point about designing BFT first, normally in FT you can punt write-contention issues to the application/user but in BFT we cannot do this because they are adversarial.)

If what you mean by "delayed write" is that transaction receipts are only generated from BCP0 and not transactors, this all makes a lot of sense... I guess you could also update for both and then clients would need to understand that the transactor receipt is not final and the BCP0 receipt is.

vyzo:
> For me this was the major lesson learned from BigChainDB: It's very difficult to apply BFT to a FT system after the fact. 

Thanks for the advice. We have already put some thought on the system, but we need to flesh out the stake model and potential coin (as transaction fuel) semantics.

> Can you please explain how either of these strategies is viable in an adversarial environment? Invalidation can be DoS vector, depending on how transactions are priced.

Transaction invalidation is more vulnerable to denial of service by spamming the system, as it is a form of optimistic concurrency. However concurrent transactions must really be within the window of transaction propagation in order to conflict and cause degradation. 
Delayed writes ensure liveness in the sense that good transactions keep making progress, but I think we have a bigger problem if we are getting spam-attacked.

> If what you mean by "delayed write" is that transaction receipts are only generated from BCP0 and not transactors, this all makes a lot of sense... I guess you could also update for both and then clients would need to understand that the transactor receipt is not final and the BCP0 receipt is.

Yeah, clients receive confirmation through the block commit protocol (which is run by the transactors).

AFDudley:
> [...]we need to flesh out the stake model and potential coin (as transaction fuel) semantics.

I think using Tendermint/Atom from within libp2p will go a long way toward simplifying the design and implementation of your stake model and coin. It's worth investigating if Sia could be integrated at this layer as well.

> Delayed writes ensure liveness in the sense that good transactions keep making progress, but I think we have a bigger problem if we are getting spam-attacked.

If there is a coin involved on a public network, attacks will happen. The financial incentives work both ways.

vyzo:
> I think using Tendermint/Atom from within libp2p will go a long way toward simplifying the design and implementation of your stake model and coin. It's worth investigating if Sia could be integrated at this layer as well.

Yeah, we are actively considering using golang and libp2p for the kittengrad implementation. 

> If there is a coin involved on a public network, attacks will happen. The financial incentives work both ways.

So true, sadly.

vyzo:
Added Radical Kittengrad draft for the new direction we are considering.

vyzo:
Merging in for reference purposes; work in this direction is currently suspended.

PR #193, opened 2016-08-08
make DynamoDatastore default to consistent reads

yusefnapora:
We're seeing what looks like a race condition on the testnet, where a newly-written record has its `canonicalInsert` event emitted by the transactor and received by the indexer, but the datastore service is returning "not found" when trying to fetch the referenced canonical.

It seems like this may be related to dynamo's default eventual consistency setting, so this PR changes the default for `DynamoDatastore` to strongly consistent reads.  You can override by setting the config property `io.mediachain.transactor.dynamo.consistentReads` to `false`.

I haven't measured the performance implications of this, so it may be that we want to keep the default and handle the inconsistencies in the clients instead.  This is especially true for IPFS, where there's no "strong consistency switch".  But I want to try deploying this branch on the testnet so I can at least determine whether this is causing the issue in the indexer.

### DIFF
--- a/transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
+++ b/transactor/src/main/scala/io/mediachain/datastore/DynamoDatastore.scala
@@ -18,6 +18,7 @@ class DynamoDatastore(config: DynamoDatastore.Config)
   val chunkSize = 1024 * 384 // 384 KB; DynamoDB has 400KB limit for item size.
                              // this item size includes field names as well as
                              // values.
+  val consistentReads = config.consistentReads
 
   val db = new AmazonDynamoDBClient()
   config.endpoint.foreach(ep => db.setEndpoint(ep))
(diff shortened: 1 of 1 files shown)


parkan:
Let me see if I understand Dynamo consistency correctly: it seems that instead of waiting for _n of m_ acknowledgement like with e.g. mongodb, consistency requirements are applied at read time, where you **ask** _n of m_ (or possibly just _m of m_) nodes and linearize results afterwards, every single read. This also appears to increase the cost (in $) of the reads by 2x, which sort of makes sense -- I would imagine the latency is increased too, because you have to wait for `max_n(...)` every time.

yusefnapora:
yeah, it looks like it does double the read cost... and I'm sure you're right about latency as well.

Maybe it is best to have the client just retry if the key isn't found

parkan:
Basically, the real (kinda) issue is that Dynamo is not an immutable/append-only store and therefore the consistency semantics are more concerned with concurrent writes than with reliably roundtripping our content-addressed immutable objects. So it's another conceptual flaw in our substitution of Dynamo for IPFS (in a way, the IPFS wait-forever behavior is the perfect complement to this behavior--both make no assumptions about when a given "write" propagates, but IPFS is infinitely optimistic and Dynamo is infinitely pessimistic, with good reason)

I'd say let's just use a manual retry for now -- incurring the consistency cost (both time and $) across _all_ reads, especially when we know that vector clock value is always the same, doesn't really make sense.

yusefnapora:
cool; I'll close this PR and work on the client end 👍

PR #194, opened 2016-08-09
update hashes in testnet.md

yusefnapora:
this slipped through the cracks after the tesnet reset; the hashes in here should all resolve now :)

### DIFF
--- a/docs/testnet.md
+++ b/docs/testnet.md
@@ -95,16 +95,21 @@ Daemon is ready
 You should then be able to immediately pull down the statement chain for an artefact:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/testnet.md
```diff
@@ -137,10 +154,10 @@ $ mediachain get QmbSMhk4EBH7SN2W4EMXuytUVYarZd2gbLdDobeJRhbU6X
 This resolves the chain head pointer, retrieves the parent cells and folds over them to give a complete
 metadata representation as some nice JSON. Straightforward (note that this object uses an older translator version format).
```
parkan: remove "(note that..." and I think we're good

yusefnapora: ah, good catch, thanks

Issue #104, opened 2016-05-27
🔮 composite/instance-specific identity


parkan:
Noting down some thoughts after @autoencoder's lead:

The current design uses a relatively conventional notion of "identity" for both artefacts and entities, which means a real canonical, platonic form, incorruptible representation of each Work exists, and when merges happen (and metadata in general is added), they happen against this one identity. One downside is that unmerges can only go back to the point of the merge; there is also no obvious way to model different notions of "sameness" with this design. 

Another approach may be to adopt a situational, contextual identity where data is added, whenever possible, to a concrete representation of the Work, potentially without bound in granularity: maybe as specific as a particular playback of a song, a print in a specific gallery, etc. These aspects of a Work's identity can then be merged at read time, depending on the query's needs. 

This can potentially be implemented with an overlay layer on top of the chains, but the multi-chain abstraction itself becomes less suitable for a system consisting of many small objects. Needs more thought.

PR #199, opened 2016-08-23
RFC4: New Directions in Mediachain [WIP]

vyzo:
The first two parts have been fleshed out enough to garner some review.
Protocol outline still WIP.


rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,450 @@
...
+And we want to allow the system to grow to very large datasets (billions
+of objects) supporting bulk and firehose ingestion models.
+
+Fundamentally, what we need is a map of media identifiers to sets
+of statements that allows us to track and relate media objects.
+This has the hallmarks of a CRDT data structure, which allows a system to
+achieve eventually consistent state without the need for consensus. 
+
+So what we want to build is a distributed database that supports
+upserts, connecting statements to one or more domain specific identifiers.
```
parkan: > jesse: "define upsert"

Upsert: an idempotent insert/update operation that does not require pre-querying for existing records, and does not rely on knowing the state of the system

parkan: > jesse: how has thinking evolved re: identifiers?

[this probably deserves its own section]

We are choosing to move from opaque/mediachain-issued identifiers to well-known/domain-specific identifiers (let's call them WKIs here) (e.g. [moma:190935](http://www.moma.org/collection/works/190935?locale=en)) with the following motivations:
- it reduces burden of authoritative ID issuance (e.g. DOI) as well as authoritatively asserting canonical-instance relationships
- most works we are dealing with at this stage will already have an ID established by previous use (see above)
- mapping from a set of WKIs to a set of statements is an elegant way of granularly and nondestructively expressing [sameness/merges](https://github.com/mediachain/mediachain/issues/104), with potentially different "sameness" perspectives represented through namespace relationships (e.g. Clarifai can publish a namespace containing only merges based on visual similarity)
- WKI-based provider records can be reasonably implemented as distributed, freeing orgs from relying on indexers for simple mappings

Works that _are_ mediachain-first and require an identifier will receive one from a ticket server/blockchain implementation (TBD). This ID will be treated the same way as any other.

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,458 @@
...
+an identifier. If the system has no prior knowledge of the identifier,
+then it creates a new entry in the identifier mapping with a singleton
+set containing the object.
+
+The lookup method can be retained as a basic read operation, with a
+change at its return value. Since there are no chained updates any
+more, the method returns the known set of updates relating to an
+identifier. 
+
+### Namespace Partitioning
```
parkan, line 92: Should look at how blockstack implements their partition rules

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,448 @@
...
+fashion.
+
+The metadata associated with statements is stored in IPLD, with the
+seeding initially supplied by sources. As statements propagate in the
+network, peers may opportunistically reseed metadata in order to aid
+availability and distribution. More specialized peers can provide
+persistent seeding and achiving services for some namespaces, perhaps
+even for a fee.
+
+### Peer Roles
```
parkan, line 184: Thoughts on this section from discussion with @jessewalden: it's quite long, but doesn't provide a ton of information on the system/protocol itself and actually raises more questions of motivation, division of responsibility, etc. Let's split it off into its own document with detailed and concrete examples, and focus on closer to the wire aspects of the protocol here: base roles can be reduced to read, write, aggregate and a separate dimension of rate/regularity.

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,328 @@
...
+providers to maintain control of their own data.
+
+In order to implement this model, each participating peer has an
+associated identity corresponding to a public key. The identities
+are intended to be long-lived and eventually have an attached reputation
+score.
+
+All statements published in the system are signed with the key of the
+source of the statement. In order to publish in a moderated namespace,
+the originating peer must present a certificate which allows it to
```
parkan: I would abstract the specific discussion of certificates/PKI at this point, I think that belongs in the more detailed design and possibly muddies the picture here

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,328 @@
...
+
+In order to reflect the natural organization of data according to institution, industry, topic, etc, we introduce the concept of namespaces, which immediately give us several advantages. In terms of scaling, the partitioning must be semantic so as to allow peers to commit resources only in topics they are interested in. In terms of administrative control, the partitioning must be hierarchical so that we can easily delegate management and moderation to organizational stake holders, see "Moderation and Publishing Model" below.
+
+To satisfy these requirements, we've chosen to adopt a hierarchical structure similar to unix paths. Leaves in the hierarchy map to specific datasets and provide the primary entry points for writes, and namespaces higher in the hierarchy aggregate namespaces below and provide a progressively expanded read view of the data structure. Much like symlinks with unix paths, alternate views of the hierarchy are possible, namespaces may have multiple parents, and so on. This creates a semantically rich and expressive framework for querying and composition, see "Namespace-based Queries and Aggregation" below.
+
+### Moderation and Publishing Model
```
parkan, line 112: Would it be correct to say that moderated namespaces assume non-byzantine behavior? I don't think that's the actual assumption but I'm not sure how to put it more precisely (trying to contrast to "open" namespaces which may have a coin fee or similar to write)

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,330 @@
...
+## A Heterogeneous Network of Cooperative Peers
+
+The distributed database is maintained by a heterogeneous peer-to-peer
+network. Peers are operated by different organizations and
+individuals, and contribute different resources to the network.
+
+Each peer maintains some parts of the dataset and has a limited view
+of the mediachain. Peers synchronize their state with peer-to-peer
+interactions, and converge their view of the data structure by
+exchanging published statements.
```
parkan, line 167: jesse correctly feels that this probably needs to be expanded upon

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,330 @@
...
+
+Statements can be propagated in the network via push, either with
+direct messages or through streaming. For stream propagation, stable
+peers form pubsub overlays for their namespaces and emit published
+statements to their subscribers. Equally well, peers can poll their
+peers for recent updates, allowing them to operate in a disconnected
+fashion.
```
parkan, line 174: I think we need to work out the details here (and in other places that deal with notion of "online" peers). The radically eventually consistent approach we are taking has very low liveness requirements, I think we can arrive at a very robust model if we follow that path thoroughly.

parkan:
Did a bunch of editing in places where I had strong views, left comments/questions where it's less clear. Ended up not hard wrapping lines and using inline MD links -- sorry! We can fix the formatting tomorrow.

Overall, definitely the right direction! Needs more attention in these areas:
- indexes/our version of provider records: my thinking here is very blurry, maybe you have a clearer view (but I am sure a clean solution exists)
- overall consistency/liveness treatment: doesn't have to be super in depth but we need to talk about it more systematically
- brief discussion of possible governance/moderation of open namespaces: nothing super concrete, just "maybe there's a coin"
- actual querying mechanics

rfc/mediachain-rfc-4.md
vyzo: I think we need to at least outline, as permissions are an essential part of the protocol. I don't intend to go into painstaking detail, just enough to support the narrative of statement publication.

rfc/mediachain-rfc-4.md
```diff
@@ -28,26 +28,34 @@ This has the hallmarks of a CRDT data structure, which allows a system to
...
+ by previous use
+* mapping from a set of WKIs to a set of statements is an elegant way of
+ granularly and nondestructively expressing sameness/merges, with potentially
+ different "sameness" perspectives represented through namespace relationships
+ (e.g. Clarifai can publish a namespace containing only merges based on visual similarity)
+* WKI-based provider records can be reasonably implemented as distributed,
+ freeing orgs from relying on indexers for simple mappings
+
+Works that are mediachain-first and require an identifier will receive one
+from a ticket server/blockchain implementation (TBD). This ID will be
```
vyzo: I don't think we need a ticket server, source can generate their own identifiers perhaps by prefixing by peer id.

vyzo:
Ok, will add some more stuff about the database semantics.

Also, I think we need a bit more about the peer roles in the main document, will expand.

vyzo:
Also the 'namespace querying and aggregation' section doesn't quite fit the narrative the way it is presented.

vyzo:
Actually it seems like the section just has the wrong section level.

rfc/mediachain-rfc-4.md
vyzo: I think we want to use a combination of push and pull. 

Always connected peers setup pubsub overlays for online update propagation. We probably want a gossip protocol instead of multicast trees for robustness and simplicity reasons (at the cost of some efficiency).

Less connected clients buffer their updates for pushing when they come back online, at which point they can publish them to the pubsub overlay. For receiving updates they poll other peers who are online.

rfc/mediachain-rfc-4.md
vyzo: I will add a couple of subsections to expand later in the section.

rfc/mediachain-rfc-4.md
vyzo: Actually, we are not overly concerned about byzantine behavior. 
The system is very loosely coupled, so other than spamming a misbehaving peer can't do much damage. There is no database corruption danger, and spam can be dealt with certificate revocation.

rfc/mediachain-rfc-4.md
parkan: Hmm I'm a bit confused about the relationship of gossip and pubsub in your description

vyzo: It's a way to implement the broadcast primitive for pubsub; it's less efficient than multicast trees obviously, as there are duplicate transmissions, but it is also a lot more robust and simple to implement.

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,475 @@
...
+signature and associated certificates independently.
+
+### Metadata Ingestion
+
+A fundamental use case for the system is the ability to ingest
+pre-existing datasets from authoritative sources in bulk.
+
+The ingestion process first requires a translation step, where
+metadata records from the existing dataset are translated to
+structured metadata objects for the mediachain. The next step is to
```
jessewalden, line 253: Modular translation out of scope, or can we make mention of it here?

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,475 @@
...
+them together with their metadata as archives. This allows the system
+to push high volumes of data without overloading the network.
+
+### Metadata Persistence
+
+The flow of metadata in the system follows the path from sources to
+readers through IPLD. When a reader receives a statement and adds it
+to its local store, it also fetches the associated metadata. Client
+readers are not expected to reseed the object in IPLD, especially if
+they operate in a disconnected fashion.
```
parkan, line 294: Should probably adjust the language here, IPLD is the format but not the exchange process, that's still IPFS.

parkan:
Moving forward! 👍

yusefnapora:
Yeah, this is shaping up well :)

I've got a few questions that I'll just dump here....
- The contents of a statement (its IPLD-referenced `object`) is a state-based CRDT set?  Or possibly a riak-style Map built on a set?  It seems like op-based CRDTs are infeasible because we can't guarantee delivery of all messages to all replicas.
- Building on that, is a statement expected to include some kind of "context" that represents the "truth" from that peer's point of view at publication time?
- For large bulk publications where you're distributing an archive of statements + source metadata, how does a peer know which archive contains the statements they care about?  How do they get a single record out of the archive without having to fetch the whole thing?
- How do peer identities map to human users?  

I've got some other thoughts on the various peer roles that I'm still trying to get together; I'll chime in with those soon

parkan:
@yusefnapora 

> The contents of a statement (it's IPLD-referenced object) is a state-based CRDT set? Or possibly a riak-style Map built on a set? It seems like op-based CRDTs are infeasible because we can't guarantee delivery of all messages to all replicas.

This was my original thinking, but I wonder if we might have monotonicity guarantees strong enough to not even need this except in the "unpublish" case? (i.e. CALM over CRDT)

> Building on that, is a statement expected to include some kind of "context" that represents the "truth" from that peer's point of view at publication time?

I'm trying to think whether this is meaningful or not

> For large bulk publications where you're distributing an archive of statements + source metadata, how does a peer know which archive contains the statements they care about? How do they get a single record out of the archive without having to fetch the whole thing?

Had similar questions regarding the batch signing -- it's a nice optimization but it does mean that the whole block (or at least the list of hashes) has to be retrieved to verify sig. I wonder if there is a clever data structure workaround for that? Outside of signing, I think the archive structure should be only thought of as a transport envelope and the individual statements can still be addressed.

> How do peer identities map to human users?

giphy-4

yusefnapora:
I guess the presence of  "causal context" (e.g vector clocks / version vectors) depends on how we're actually implementing the statements.  I'm reading through the bloom / CALM stuff in more detail now; it seems premised on there being some form of strong consistency (paxos, etc) for the parts of the computation that are not monotonic.  Although if we can fit everything into the CALM zone, that'd be great :)

I think I'm drawn to state-based CRDTs because they seem somewhat resilient to a statement completely disappearing from the network.  Especially since the ipfs "pinning economy" is still unclear,  I think we should assume that statements can drop out of existence despite our best efforts at caching, archiving, etc.  Which seems to rule out an op-log style approach, although I might be missing something that could make it work.

rfc/mediachain-rfc-4.md
vyzo: I think the details of the translation are a bit out of scope, as they pertain to the meta object schema etc.

rfc/mediachain-rfc-4.md
vyzo: Good point, I'll update.

vyzo:
> > The contents of a statement (it's IPLD-referenced object) is a state-based CRDT set? Or possibly a riak-style Map built on a set? It seems like op-based CRDTs are infeasible because we can't guarantee delivery of all messages to all replicas.
> 
> This was my original thinking, but I wonder if we might have monotonicity guarantees strong enough to not even need this except in the "unpublish" case? (i.e. CALM over CRDT)

I don't think we can force unpublish, except perhaps as a gc hint.

> For large bulk publications where you're distributing an archive of statements + source metadata, how does a peer know which archive contains the statements they care about? How do they get a single record out of the archive without having to fetch the whole thing?

The idea is that archives will get published and find their way to indexers and readers, who can unpack them and then answer queries based on the content. This allows the extraction of individual statements, but signature verification will still require the entire block. The block don't have to be monstrously big however (ie 100MM signatures in one go), the batches can have a reasonable number of statements (say 1k).

> How do peer identities map to human users? 

Humans can attach a name. email, etc on their keys, much like PGP keys.

rfc/mediachain-rfc-4.md
```diff
@@ -555,16 +555,88 @@ revocation has taken effect. Peers can also periodically run garbage
...
+compound-statement-body = [<simple-statement-body>, ...]
+
+republished-statement-body = [<Statement>, ...]
+
+statement-id = <ID>:<timestamp>:<nonce>
+```
+
 #### Archives
 
+Archives are gzipped tarballs with the following structure:
```
parkan, line 434: I wonder if tarball is optimal for this

rfc/mediachain-rfc-4.md
```diff
@@ -555,16 +555,88 @@ revocation has taken effect. Peers can also periodically run garbage
...
+  ...
+ ipld/
+  ...
+```
+
+Archives have an id which is derived as a statement identifier from
+the source of the archive. The `stmt` directory contains all the
+statements pertaining to objects in the archive, in files named after
+the statement id they contain. The `ipld` directory is structured as a
+hash tree and contains all metadata objects, public keys, and
```
parkan: seems like what we really have is an index of statements + IPLD hashtree, doesn't sound like a linear tarball to me

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,653 @@
...
+
+A basic indexing scheme, which performs only statement envelope processing
+and doesn't parse the metadata, allows retrieval with the following
+criteria:
+
+- Query by WKI, publisher or peer id, statement id, object id
+- Query by namespace(s)
+- Query by timestamp
+
+This baseline index, which can be easily implemented using an embedded
```
parkan, line 327: I think we may be able to use some interesting distributed data structures to manifest this index, rather than simply having it be peer-local -- @jbenet had some thoughts on this

rfc/mediachain-rfc-4.md
```diff
@@ -0,0 +1,653 @@
...
+synchronization. It also provides a low level interface for advanced
+users, who are free to develop and share more advanced indexing schemes
+that parse object content.
+
+Queries do not have to be limited to one's local view however. For
+instance, a user may want to query joining on namespaces it is not
+currently following.  This could be because the user is exploring the
+namespace or because the particular namespace has a larger index that
+aggregates multiple namespaces. This can be accomplished by directly
+querying other peers or by utilizing more specialized indexers.
```
parkan, line 338: Likewise here I wonder if we can have hierarchical, mergeable indexes

jbenet:
( On my stack. I'll give feedback on this today or tmrw, before our call. )

rfc/mediachain-rfc-4.md
vyzo: Maybe not, but it's the simplest option to start with.

rfc/mediachain-rfc-4.md
vyzo: The idea is that the ipld directory is split to multiple directories per hash entry, so as to avoid having too many files in leave directories. Not sure how to describe it better than a 'hash tree'.

rfc/mediachain-rfc-4.md
vyzo: I am a little weary of adding fancy distributed data structures at this level -- the local store is a simple and robust solution. 

Note that we already have a distributed data structure overall, these are just the individual components.

rfc/mediachain-rfc-4.md
vyzo: I think we probably want to do queries that join across multiple nodes, but the join can be implemented purely in the client side with stacked queries.

vyzo:
Performed surgery to remove certificates from the system, and move to a reputation/endorsement moderated publication model.

parkan:
@jbenet feel free to throw up any inline comments we haven't covered. I'll try to edit this a bit to reflect the clarifications we arrived at -- I think it's mostly a matter of phrasing and focus.

vyzo:
Added docs/concat.md which describes the MVP experiment for the PME [Pre-critical Mass Era]

docs/concat.md
```diff
@@ -0,0 +1,174 @@
+# ConCat: A System Design for Mediachain PME [Pre-critical Mass Era]
+
+## Architecture
+
+The system follows the RSS model, whereby individual peers maintain
```
parkan: maybe "inspired by" rather than "follows"

docs/concat.md
```diff
@@ -0,0 +1,174 @@
...
+can be directly connected to real world entities and carry their
+associated reputation with them.
+
+The public identification allows directory servers to provide the next
+stratum of discovery, by virtue of the Contrib directory. This
+directory lists only verified nodes and allows discovery based on
+operator identities (eg organizations, social media handles, keybase
+user names and so on).
+
+The final stratum of discovery, is the Glam directory. In this space,
```
parkan, line 31: agreed w/denis that "Glam" is not clear here, though it is amusing 😄

docs/concat.md
vyzo: ok, will fix.

vyzo:
Merging in, as it has reached a stable point.

Issue #200, opened 2016-08-24
[DISCUSSION] API prototype + network prototype split


parkan:
One downside to our approach taken with Phase 1, where we attempted to meet a certain level of compromise in every part of the system, was that certain realizations about API and pipeline properties came relatively late. Given that we're changing the API fairly substantially with this iteration, it may be wise to separate the working prototypes into a fully centralized _API prototype_ and a network prototype/testbed.

I am thinking something like this:
### API prototype
- Builds on existing work with indexer pipeline + P1 prototype
- SImulates 1 optimally connected peer in topmost namespace
- Backed by relatively flexible datastore that can model our index structures (though postgresql is probably fine tbh)
- Wiring up to medium-indexer
- Prototyping [user stories](https://github.com/mediachain/mediachain/blob/7ba0762411fdb47a91a0c18c3346d705bc3c62c7/rfc/mediachain-rfc-4-roles.md)
- Query testbed
- Establish performance requirements for distributed net
- Can iterate much faster and w/o disruption from unstable network

@autoencoder and @parkan would be responsible, with help from @denisnazarov 
### Network prototype
- Builds on top of libp2p
- Prototyping PKI, pubsub, discovery, etc doesn't require any cooperation from application level
- Many other pieces can be built in relative isolation
- Build into API shape developed from iteration on API proto
- Will need real data to test with, can capture that as dumps from API proto

@vyzo and @yusefnapora would be responsible

What do you think about this arrangement?

parkan:
input from @yusefnapora @autoencoder @denisnazarov @jessewalden welcome

parkan:
(note that the previously clear distinction between internal and external API is now much less so, so separation may be a bit tricky -- welcome any thoughts on this too)

autoencoder:
This mock backing, to be replaced with the real version as it's developed (or kept around for testing needs), and having the API drafted up from the beginning and working backward to the implementation -- exactly what I was thinking. Great!

Both the API sketch & protocol sketch doc are being drafted in parallel, since the API design can drastically change the protocol, right?

I'll add that the API sketch should at least include these 3 things:
- API calls and arguments.
- Example minimal inputs and outputs for each API call, and possibly for any groups of calls needed to perform a single conceptual action. 
[...]

PR #202, opened 2016-08-29
[discussion] notes for next-gen surface api

yusefnapora:
Putting some notes here related to #200 for easier review & collaboration.

### DIFF
--- a/docs/future-api.md
+++ b/docs/future-api.md
@@ -1,0 +1,129 @@
+## mediachain.next api notes
+
+Everything here is in flux and subject to change; this is just a sketch. 
+The idea is to give an overview of the high-level api surface of an "online" node, that is, 
+one that actively contributes to the network and is expected to remain connected to the network.
+
+All the code examples below are in javascript, with flow-style annotations for return types.
+
+### Peer ids & local state
+A peer has a unique id, which is the hash of their public key.  They may also have a collection of 
+certificates, signed by moderators of the namespaces they're authorized to participate in.
+
+The peer-local state includes the peer's id, certificate store, private key, etc.  It also includes the 
+index of connected peers, which allows us to route operations to peers with the correct roles.
+
+For the examples below, the peer-local state is assumed to exist in an accesible place 
+(like an object instance variable, etc), so it's omitted from the function arguments.
+
+### Domain-specific ids (Well-Known-Identifiers)
+The RFC-4 design uses the term Well-Known-Identifier (WKI) to refer to an existing domain specific id, 
+for example `nypl:510d47e2-ef28-a3d9-e040-e00a18064a99` which refers to an object in the New York Public Library collection.
+
+**question:** is there a shared directory of WKI prefixes? How does a user know to use `getty` instead 
+of e.g. `gettyimages`, or that `getty` doesn't refer to http://www.getty.edu/?
+
+### Connect / peer discovery
+Ideally a single `connect()` method that optionally allows you to specify the address of a bootstrap peer / 
+directory server. If not provided, picks a peer from a hard-coded set built into the app. 
+
+```javascript
+function connect(options): Promise<Unit> {
+  let {bootstrapAddress} = options
+  // ...
+}
+```
+
+*mock design*: This is a noop, as we're mocking a fully connected node with a direct route to all (virtual) peers.
+
+*p2p design*: Contacts a directory server to update the peer index and register any services we're capable of providing. 
+May push updates that we've spooled locally to any peers that were previously subscribed to us.  
+May register for periodic updates from the director server.
+
+### Publication
+A node can publish to any leaf namespace that it has write permission for.  If a namespace is not specified, statements 
+are published to the user's personal namespace, which corresponds to the hash of their peer id, e.g. `universe.peers.QmF00123` 
+
+**questions:** should we accept a list of namespaces, to publish to multiple namespaces at once?
+
+```javascript
+function publish(statement): Promise<StatementID> { ... }
+
+publish({
+  wkis: ["getty:530122144"],
+  meta:{
+    "orientation": "Horizontal",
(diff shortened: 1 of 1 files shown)


docs/future-api.md
```diff
@@ -0,0 +1,129 @@
...
+index of connected peers, which allows us to route operations to peers with the correct roles.
+
+For the examples below, the peer-local state is assumed to exist in an accesible place 
+(like an object instance variable, etc), so it's omitted from the function arguments.
+
+### Domain-specific ids (Well-Known-Identifiers)
+The RFC-4 design uses the term Well-Known-Identifier (WKI) to refer to an existing domain specific id, 
+for example `nypl:510d47e2-ef28-a3d9-e040-e00a18064a99` which refers to an object in the New York Public Library collection.
+
+**question:** is there a shared directory of WKI prefixes? How does a user know to use `getty` instead
```
parkan, line 23: This is a good question, and one of the synchronization points. I think there are a few rules of thumb (e.g. namespace name and related prefix should be the same), but I don't have a clear picture of how this will work at scale.

Converging two prefixes (e.g. "themet" + "metropolitan") is exactly a merge operation, so it's pretty cheap (+1 statement), but separating ambiguous prefixes is trickier, so we probably want to avoid it at write time as much as possible, though we have to assume that it _will_ happen at some rate. The question then becomes whether we want to try to be proscriptive about this at all, or try to synthesize preferred prefixes from usage.

My intuition is that we have to draw the line somewhere, and that this is a pretty good place for the line: using an incorrect prefix should be considered as wrong as other types of data errors.

docs/future-api.md
```diff
@@ -0,0 +1,129 @@
...
+
+For the examples below, the peer-local state is assumed to exist in an accesible place 
+(like an object instance variable, etc), so it's omitted from the function arguments.
+
+### Domain-specific ids (Well-Known-Identifiers)
+The RFC-4 design uses the term Well-Known-Identifier (WKI) to refer to an existing domain specific id, 
+for example `nypl:510d47e2-ef28-a3d9-e040-e00a18064a99` which refers to an object in the New York Public Library collection.
+
+**question:** is there a shared directory of WKI prefixes? How does a user know to use `getty` instead 
+of e.g. `gettyimages`, or that `getty` doesn't refer to http://www.getty.edu/?
```
parkan, line 24: The getty family of... things (getty images, getty museum, getty villa, getty research institute, getty foundation, ...) is a great case study for this, actually. I wonder if they have an internal guide for what to call what.

docs/future-api.md
```diff
@@ -0,0 +1,129 @@
...
+
+*mock design*: This is a noop, as we're mocking a fully connected node with a direct route to all (virtual) peers.
+
+*p2p design*: Contacts a directory server to update the peer index and register any services we're capable of providing. 
+May push updates that we've spooled locally to any peers that were previously subscribed to us.  
+May register for periodic updates from the director server.
+
+### Publication
+A node can publish to any leaf namespace that it has write permission for.  If a namespace is not specified, statements 
+are published to the user's personal namespace, which corresponds to the hash of their peer id, e.g. `universe.peers.QmF00123`
```
parkan, line 45: let's draw some diagrams illustrating the various publish/republish/push/pull models possible here

docs/future-api.md
```diff
@@ -0,0 +1,129 @@
...
+
+*p2p design*: Contacts a directory server to update the peer index and register any services we're capable of providing. 
+May push updates that we've spooled locally to any peers that were previously subscribed to us.  
+May register for periodic updates from the director server.
+
+### Publication
+A node can publish to any leaf namespace that it has write permission for.  If a namespace is not specified, statements 
+are published to the user's personal namespace, which corresponds to the hash of their peer id, e.g. `universe.peers.QmF00123` 
+
+**questions:** should we accept a list of namespaces, to publish to multiple namespaces at once?
```
parkan, line 47: at API level, sure, but let's keep it a list of 1 for now to minimize concerns like atomicity etc

docs/future-api.md
parkan, line 23: From a UX perspective, we should have a special index of known prefixes and illustrative examples for each

docs/future-api.md
```diff
@@ -0,0 +1,129 @@
...
+#### Bulk publication
+The bulk publication api is very similar to the single-statement api, but accepts a list of statement objects.
+
+```javascript
+function publish_bulk(statements): Promise<Array<StatementID>> { ... }
+```
+
+*mock design:* no significant difference from single-statement implementation.
+
+*p2p design:* bulk publication differs in implementation.  Statements will be signed in batches and distributed
```
parkan, line 77: did we end up discussing the batch validation etc procedures while I've been gone?

docs/future-api.md
```diff
@@ -0,0 +1,129 @@
...
+*p2p design:* bulk publication differs in implementation.  Statements will be signed in batches and distributed 
+in archives for efficiency.  This may complicate the implementation for reads, when a peer requests a small 
+subset of the archive.
+
+
+#### Note about ID extraction, translation, etc.
+The `publish` functions above assume a pre-processing step which maps ids to the global namespace of Well-Known-Identifiers.  
+For example, the getty image id `530122144` is mapped to `getty:530122144`.
+
+The contents of the `meta` field may also need some amount of translation, but the specifics of that are still not fully worked out.
```
parkan, line 86: My thinking on this is something like:
- move away from the redistributable translators as default (though they may still be packaged in this way in a parallel system)
- translator field version should be semantic so tools can know if they can work with a format (e.g. I can ask for statements in `autoencoder-indexer>0.2` format)

parkan:
Overall good direction, I think this can definitely benefit from @autoencoder questions as prompts for what to focus on (when you have time, tg)

